PR #1, opened 2017-06-30
Fix Readme build step

yanndegat:
build step still references github.com/hashicorp


stack72:
LGTM! thanks for this @yanndegat :)

PR #7, opened 2017-08-22
Docs: fix copy/paste error

jgoldschrafe:
`vrack_publiccloud_attachment`'s `project_id` defaults to `OVH_PROJECT_ID`, not `OVH_VRACK_ID`.


jgoldschrafe:
This PR is for docs correctness, but I noticed that OVH_PROJECT_ID is an environment variable that only exists for use within Terraform. Should it be looking at the standard variables `OS_PROJECT_ID`/`OS_TENANT_ID` instead, or is the idea to avoid the assumption that the "cloud project" ID is an OpenStack project/tenant ID?

PR #13, opened 2018-01-26
Public cloud monthly billing

Maelstrom96:
Hi,

I know this PR is missing some tests but I wasn't sure how to implement them since each test costs some dollars since there's no API sandbox offered by OVH. Even the cheapest instance is 5.99$ monthly and is getting billed to your credit balance.

I'm currently using this and it seems to work quite well, I know that's monthly billing isn't really a _resource_ but in order to be logical, I would have had to recreate the instance_v2 resources from OpenStack.

We could maybe still implement it in another way but it works for what I needed it for and wanted to see if you were going to implement something similar. If you have another route that you would like to go with for this implementation, let me know and I'll see if I can re-implement it the way that you are seeing this.

Thanks

### DIFF
--- a/ovh/provider.go
+++ b/ovh/provider.go
@@ -53,6 +53,7 @@ func Provider() terraform.ResourceProvider {
 			"ovh_publiccloud_user":                   resourcePublicCloudUser(),
 			"ovh_vrack_publiccloud_attachment":       resourceVRackPublicCloudAttachment(),
 			"ovh_domain_zone_record":                 resourceOVHDomainZoneRecord(),
+			"ovh_publiccloud_billing_monthly":        resourcePublicCloudBillingMonthly(),
 		},
 
 		ConfigureFunc: configureProvider,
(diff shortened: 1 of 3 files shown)


yanndegat:
hi @Maelstrom96 

thank you for your work. it really great to have ppl onboard with us.
i don't think we can accept your PR, yet i clearly understand that this feature would be great. 

2 things: 
- it's not possible to have a resource which has a "delete" impl that "doesn't delete"
- we think, for many reasons, that the ovh provider should have its own "ovh_publiccloud_instance" resource, even if the impl only proxyfies to the openstack provider. 

That said, it means that once we have the instance resource within the ovh provider, we would then be able to "mix" features from ovh+openstack. Thus your feature could be impl wihtin this resource, and the "delete" issue will not be an issue anymore.

i dont have an ETA for this but its clearly something we have in mind.

so i suggest we open an issue related to this work and mention the monthly billing feature to keep track of this.

what do you think?

PR #16, opened 2018-03-08
Fix doc on zone record resource

eliecharra:
Case issues in documentation


yanndegat:
hi @eliecharra 

thanks a lot.

PR #18, opened 2018-04-04
vendor: update github.com/ovh/go-ovh/ovh

thbkrkr:
Hi,

The vendorized lib github.com/ovh/go-ovh/ovh has not been updated for a while.

I would like to have the fix (https://github.com/ovh/go-ovh/commit/34b9ffdf98692f9e9b57724d180549f88725b605).

Therefore can we update the lib?

Thanks.


yanndegat:
thanks a lot @thbkrkr

PR #22, opened 2018-05-02
Vps resource

gaetanfl:
This PR create a first implentation of the vps resource for terraform.
At the moment creation is not supported, only importing. Basic modification is possible.
Only import is tested at the moment, due to the lack of creation support and the fact I don't know how to persist a state from an imported resource in a TestStep to another.

Ideas for creation:
* Create a data source for CreditCard Payment Mean to ensure we can do the order
* Create a data source for vps model type

Ideas for deletion:
* A two step deletion:
  1. Use state=terminating to issue the `POST /vps/{serviceName}/terminate`
  2. Pass a var with the received code to issue `POST /vps/{serviceName}/confirmTermination`
* Considered it deleted after a `POST /vps/{serviceName}/terminate`

### DIFF
--- a/ovh/resource_ovh_vps.go
+++ b/ovh/resource_ovh_vps.go
@@ -247,7 +247,7 @@ func resourceOVHVPSUpdate(d *schema.ResourceData, meta interface{}) error {
 			return fmt.Errorf("Unknown wanted state")
 		}
 		err := provider.OVHClient.Post(
-			fmt.Sprintf("/vos/%s/%s", d.Id(), command),
+			fmt.Sprintf("/vps/%s/%s", d.Id(), command),
 			nil,
 			nil,
 		)
--- a/ovh/resource_ovh_vps.go
+++ b/ovh/resource_ovh_vps.go
@@ -272,6 +272,7 @@ func OVHVPS_getType(offertype string, model_name string, model_version string) s
 	offertypeToOfferPrefix["cloud"] = "ceph-nvme"
 	offertypeToOfferPrefix["cloud-ram"] = "ceph-nvme-ram"
 	offertypeToOfferPrefix["ssd"] = "ssd"
+	offertypeToOfferPrefix["classic"] = "classic"
 	return (fmt.Sprintf("vps_%s_%s_%s", offertypeToOfferPrefix[offertype],
 		model_name,
 		model_version))
--- a/ovh/resource_ovh_vps.go
+++ b/ovh/resource_ovh_vps.go
@@ -55,10 +55,6 @@ func resourceOVHVPS() *schema.Resource {
 		},
 
 		Schema: map[string]*schema.Schema{
-			"id": {
-				Type:     schema.TypeString,
-				Computed: true,
-			},
 			"type": {
 				Type:     schema.TypeString,
 				Required: true,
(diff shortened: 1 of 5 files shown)


yanndegat:
hi @gaetanfl 

thanks a lot for your work
it great to have people working on new resources

i'll made small comments before taking time to test it and run the acc tests

ovh/resource_ovh_vps.go
```diff
@@ -0,0 +1,283 @@
...
+		Create: resourceOVHVPSCreate,
+		Read:   resourceOVHVPSRead,
+		Update: resourceOVHVPSUpdate,
+		Delete: resourceOVHVPSDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"id": {
```
yanndegat: id should be present in resources schema
this is a common mistake

ovh/resource_ovh_vps.go
```diff
@@ -0,0 +1,283 @@
...
+		&vps,
+	)
+
+	if err != nil {
+		d.SetId("")
+		return nil
+	}
+
+	d.Set("id", vps.Name)
+	d.Set("id", vps.Name)
```
yanndegat: remove both lines and use SetId func instead

ovh/resource_ovh_vps.go
```diff
@@ -0,0 +1,283 @@
...
+		}
+	}
+	return nil
+}
+
+func resourceOVHVPSDelete(d *schema.ResourceData, meta interface{}) error {
+	return fmt.Errorf("Deletion is not supported at the moment, please terminate manually and terraform state rm")
+}
+
+func OVHVPS_getType(offertype string, model_name string, model_version string) string {
```
yanndegat: this would be better if not exported outside package
(thus lowercase prefix)

yanndegat:
hi @gaetanfl 

thanks a lot for this new resource
it's great to have people working on these parts

gaetanfl:
I just updated as your comments and fixed a small typo.
Concerning a resource vs datasource it's just creation is planned but not yet done.

yanndegat:
hi @gaetanfl 

do you think you can explain the api calls workflow you'd like to setup 
to impl the create/delete missing funcs ?
because i'm not so sure we can impl by simple api calls. but maybe i've missed something

gaetanfl:
hi @yanndegat,
It can be achived by creating a cart with a post on /order/cart getting the id then assign the cart to my account and create an order for a vps with a post on /order/cart/$id/vps getting the item_id and then do some posts on /order/cart/$id/vps/$item_id/configuration to specify options
and then checkout the cart with a post on /order/cart/$id/checkout and pay the bill with a registered payment mean (creditcard) with a call on /me/order/$id/payWithRegisteredPaymentMean

gaetanfl:
@yanndegat sorry I was working on other things I completly forget to answer. We can stick to this model. It suits my needs which are to keep track of server we manage in terraform.

gaetanfl:
Since similar work was done on dedicated server with data_source I guess the best way to do it is through datasource and update resources

yanndegat:
hi @gaetanfl 

i'm sorry i have to close this PR
you have done great job but i could not get the proper resources to merge it as is.
 it is stalled for too long now & as you suggested, maybe we should prefer the way we did it for dedicated servers.

PR #29, opened 2018-05-11
fixes #27 by switching to /auth/currentCredential for client validation

### LINKED ISSUE #27: Unable to use provider with no access to `/me` api endpoint
### Expected Behavior
Having OVH Api application credentials I should be able to manage resources I have been granted access to.

### Actual Behavior
If access to API on `/me` URI was not granted the provider fails to run with

    * provider.ovh: OVH client seems to be misconfigured: "Error 403: \"This call has not been granted\""    

### Suggested solution
For verification of client connection switch from `/me` to `/auth/time` which returns unix timestamp and does not require explicit access rights granted.
While no authentication is required to access this endpoint, if provided and invalid it will still fail.

goblain:


yanndegat:
thanks again

Issue #23, opened 2018-05-10
Reflect API path in resource naming


goblain:
It would seem reasonable to keep names of resources a bit more in line with what can be found in OVH API. 
For example, considering the URI in OVH API is 
`/cloud/project/{serviceName}/network/private/{networkId}/subnet`
the resource name for private networks subnet would be
`ovh_cloud_network_private_subnet` 
instead of 
`ovh_publiccloud_private_network_subnet`

I could adapt that keeping previous naming for backward compatibility. It would also provide a clear guideline on how to name new resources.

yanndegat:
released

PR #31, opened 2018-05-14
adds new resource names following issue #23 outlines

goblain:


yanndegat:
hi @goblain 

thanks a lot
this makes sense

do you think you could also update the doc accordingly?

how could we add info on the "legacy resources" to deprecated them?

yanndegat:
hi @goblain 

thanks a lot, that's super great from you

PR #32, opened 2018-05-17
adds ovh_iploadbalancing_tcp_farm resource

goblain:


yanndegat:
hi @goblain 

that's "super dope" ! thank you very much
this resource is expected by a lot of people

we have to add acc tests along with it and complete the doc entries

do you think you could do it ?

i'll see on my side how i can deal with hashicorp & ovh so that we would have an iplb
instance dedicated for acc tests

goblain:
docs and basic test added

eliecharra:
> i'll see on my side how i can deal with hashicorp & ovh so that we would have an iplb
instance dedicated for acc tests

@yanndegat I'm thinking about adding iplb routing resources in a second time but this require Offer pack 2 (with advanced features), which is quite expensive (200€/mo) Are you aware of what kind of service hashicorp could negotiate for integration tests ?

yanndegat:
hi @eliecharra 

i actually work @ovh, i have to deal with some folks internally to see what is possible or not

eliecharra:
@yanndegat Sounds great !

Btw thanks a lot @goblain for this killer feature :+1:

yanndegat:
news is: acc test on hc's ci infra should be solved next week

until then ill try to run all this on my infra 

again thanks a lot @goblain for all the hard work

eliecharra:
> i think once we will have merge the vps, and the iplb tcp farm we will ask hc for a release

Nice :+1: 

I'll try to add a `ovh_iploadbalancing_route` afterwards.

PR #33, opened 2018-05-21
adds ovh_iploadbalancing_tcp_farm_server resource

goblain:
This commit introduces `ovh_iploadbalancing_tcp_farm_server`.

### DIFF
--- a/ovh/resource_ovh_iploadbalancing_tcp_farm_server.go
+++ b/ovh/resource_ovh_iploadbalancing_tcp_farm_server.go
@@ -1,0 +1,234 @@
+package ovh
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"net"
+
+	"github.com/hashicorp/terraform/helper/schema"
+)
+
+type IpLoadbalancingTcpFarmServer struct {
+	BackendId            int     `json:"backendId,omitempty"`
+	ServerId             int     `json:"serverId,omitempty"`
+	FarmId               int     `json:"farmId,omitempty"`
+	DisplayName          *string `json:"displayName,omitempty"`
+	Address              *string `json:"address"`
+	Cookie               *string `json:"cookie,omitempty"`
+	Port                 *int    `json:"port"`
+	ProxyProtocolVersion *string `json:"proxyProtocolVersion"`
+	Chain                *string `json:"chain"`
+	Weight               *int    `json:"weight"`
+	Probe                *bool   `json:"probe"`
+	Ssl                  *bool   `json:"ssl"`
+	Backup               *bool   `json:"backup"`
+	Status               *string `json:"status"`
+}
+
+func resourceIpLoadbalancingTcpFarmServer() *schema.Resource {
+	return &schema.Resource{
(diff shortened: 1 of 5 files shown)


yanndegat:
hi @goblain 

regarding the ovh_ip_server;
couldn't we use an ip from a private ipv4 range instead ? such as 192.168.0.1 ? or 10.0.0.10 ?
as we will add a "vrack enabled" iplb, it shouldnt be rejected by the API

goblain:
using private IP seems to yield satisfying result, so indeed no need for this env

ovh/resource_ovh_iploadbalancing_tcp_farm_server.go
```diff
@@ -0,0 +1,234 @@
...
+			},
+		},
+	}
+}
+
+func resourceIpLoadbalancingTcpFarmServerCreate(d *schema.ResourceData, meta interface{}) error {
+	config := meta.(*Config)
+
+	newBackendServer := &IpLoadbalancingTcpFarmServer{
+		DisplayName:          getNilStringPointer(d.Get("display_name").(string)),
```
yanndegat: i don't see the use of "getNilStringPointer"

AFAIK, there's no such thing elswhere in terraform providers. what are we trying to do here ?

goblain: assuming "display_name" is not set, `d.Get("display_name").(string)` returns an empty string `""`, but a string none the less. gos `json.Marshal` method correctly interprets it as an empty string rather then a *not set* string. This is why the struct uses pointers instead of string values, so they can have `nil` value. getNilStringPointer then gets the string value and turns it into a pointer to string value, that in case of `len(<string>) == 0` returns nil, so that `json.Marshal()` instead of generating `"displayName": ""` generates `"displayName": null` which OVH API expects (ie. it will complain if you will have a zero value field instead of null for attributes that can not be changed after resource creation

yanndegat: have you tried with d.Get("display_name").(*string) ?

goblain: in fact I did :) it did not work. Don't think type assertion is meant to work like that at all, and even if it did, it would have no way to distinct between nil and zero value, so instead of recovering `nil`, it would recover `*""` but as I said, it's not doing even that.

goblain: the functions I provided here do just a bit more then what you can see in Pointer retrievals for aws like : 
https://github.com/terraform-providers/terraform-provider-aws/blob/80e4ae53941c6fc87239f1d704e2dfe6a4dec3b6/aws/resource_aws_mq_broker.go#L185
https://github.com/aws/aws-sdk-go/blob/71a68668c287f9ad58d0c58c78e27f08eb39fa49/aws/convert_types.go#L5

goblain:
rebased

yanndegat:
once again, thanks a lot @goblain

PR #34, opened 2018-05-22
Payment data sources

gaetanfl:
In order to make payment for vps resource creation or any other kind of classic (non cloud) ovh products a data source for payments method would make things easier.

I've a question regarding tests, if I use two env variable with filter patterns for credit card or bank account and check there is a match, is it enough ?

### DIFF
--- a/ovh/provider.go
+++ b/ovh/provider.go
@@ -46,11 +46,11 @@ func Provider() terraform.ResourceProvider {
 			// New naming schema (issue #23)
 			"ovh_cloud_region":  dataSourcePublicCloudRegion(),
 			"ovh_cloud_regions": dataSourcePublicCloudRegions(),
+			"ovh_credit_card":   dataSourceCreditCard(),
+			"ovh_bank_account":  dataSourceBankAccount(),
 			// Legacy naming schema (new datasources should not be added here)
 			"ovh_publiccloud_region":  dataSourcePublicCloudRegion(),
 			"ovh_publiccloud_regions": dataSourcePublicCloudRegions(),
-			"ovh_credit_card":         dataSourceCreditCard(),
-			"ovh_bank_account":        dataSourceBankAccount(),
 		},
 
 		ResourcesMap: map[string]*schema.Resource{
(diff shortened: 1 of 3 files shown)


tombuildsstuff:
hey @gaetanfl 

Thanks for this PR :)

Taking a quick look at this PR it contains sensitive information (IBAN / Credit Card Numbers) which aren't encrypted in the state. Since this is the first time this kind of information has been exposed across providers - we need to chat about this internally before proceeding - please bear with us for the moment and we'll get back to you shortly.

Thanks!

gaetanfl:
Hi,
in fact IBAN and credit card numbers are truncated you have just the beginning and the end which shouldn't be a problem

yanndegat:
hi @tombuildsstuff  & @gaetanfl 

actually, i'm also checking internally @ovh if this kind of info isnt to sensible to be exposed in a terraform state file.

i'm not so sure we want to take this risk.

yanndegat:
hi @gaetanfl 

i've checked internally. there should be no problem outputting these values as they are returned by the API
yet, terraform is an infrastructure tool and there are sensitive info not useful to bootstrap infrastructure in the computed attributes. 

Do you think we could export only the internal id return by the API if the datasource is found ? 
or do you need other attributes ?

gaetanfl:
I guess we could keep for bank account:
* Description
* Default
* State
* Id

and for Credit Card:
* Description
* Expiration
* Default
* State
* Id

Will that be ok for you ?

gaetanfl:
I will just remove it, if someone need it it won't be difficult to implement

jkohrman:
Hi there @gaetanfl - This is looking much better, thanks!

Quick question for you since I'm not familiar with this particular OVH API.  What is the expected value of the `State` attribute?

ovh/provider.go
```diff
@@ -46,6 +46,8 @@ func Provider() terraform.ResourceProvider {
 			// New naming schema (issue #23)
 			"ovh_cloud_region":  dataSourcePublicCloudRegion(),
 			"ovh_cloud_regions": dataSourcePublicCloudRegions(),
+			"ovh_credit_card":   dataSourceCreditCard(),
```
yanndegat: could we map the ovh api by prefixing datasources by "ovh_me_" ?
pls ? and rename the go functions accordingly ?

thanks a lot

PR #54, opened 2018-09-06
r/ovh_ip_reverse: new resource

pkuhner:
This PR adds an IP reverse resource allowing to set the reverse DNS of an instance's IP.

Example resource:

```
resource "ovh_ip_reverse" "my_ip_reverse" {
    ipreverse = "192.0.2.42"
    reverse = "example.com"
}
```

To do:
- Add tests
- Add doc
- Add a retry mechanism on resource creation to avoid `Error 400: "example.com. does not resolve to 192.0.2.42" when creating the A record and the IP reverse in the same terraform plan

### DIFF
--- a/ovh/resource_ovh_ip_reverse.go
+++ b/ovh/resource_ovh_ip_reverse.go
@@ -1,0 +1,176 @@
+package ovh
+
+import (
+	"fmt"
+	"log"
+	"net"
+	"strings"
+
+	"github.com/hashicorp/terraform/helper/schema"
+
+	"github.com/ovh/go-ovh/ovh"
+)
+
+type OvhIpReverse struct {
+	IpReverse string `json:"ipReverse"`
+	Reverse   string `json:"reverse"`
+}
+
+func resourceOvhIpReverse() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceOvhIpReverseCreate,
+		Read:   resourceOvhIpReverseRead,
+		Update: resourceOvhIpReverseUpdate,
+		Delete: resourceOvhIpReverseDelete,
+
+		Schema: map[string]*schema.Schema{
+			"ip": {
+				Type:     schema.TypeString,
+				Required: true,
+				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+					err := validateIpBlock(v.(string))
+					if err != nil {
+						errors = append(errors, err)
+					}
+					return
+				},
+			},
+			"ipreverse": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
+					err := validateIp(v.(string))
+					if err != nil {
+						errors = append(errors, err)
+					}
+					return
+				},
+			},
+			"reverse": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+		},
+	}
+}
+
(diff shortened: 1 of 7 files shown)


pkuhner:
- Removed ForceNew attributes and added Update function.
- Added a (dirty) retry mechanism on Create

ovh/resource_ovh_ip_reverse.go
```diff
@@ -0,0 +1,154 @@
...
+		d.Set("ip", newIp)
+	}
+
+	log.Printf("[DEBUG] OVH IP Reverse create configuration: %#v", newReverse)
+
+	resultReverse := OvhIpReverse{}
+
+	var err error
+
+	for i := 0; i < 5; i++ {
```
yanndegat: why?
i think we should leave this "retry management" to the end user. if it fails, the end user can retry to apply the plan.

ovh/resource_ovh_ip_reverse.go
```diff
@@ -0,0 +1,154 @@
...
+	provider := meta.(*Config)
+
+	// Create the new reverse
+	newReverse := &OvhIpReverse{
+		IpReverse: d.Get("ipreverse").(string),
+		Reverse:   d.Get("reverse").(string),
+	}
+
+	newIp, ok := d.GetOk("ip")
+	if !ok {
```
yanndegat: !ok || newIp == ""

ovh/resource_ovh_ip_reverse.go
```diff
@@ -0,0 +1,154 @@
...
+func resourceOvhIpReverse() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceOvhIpReverseCreate,
+		Read:   resourceOvhIpReverseRead,
+		Update: resourceOvhIpReverseUpdate,
+		Delete: resourceOvhIpReverseDelete,
+
+		Schema: map[string]*schema.Schema{
+			"ip": {
+				Type:     schema.TypeString,
```
yanndegat: ip shoud have a validate function which checks the the string is formatted as an ipv4 block

ovh/resource_ovh_ip_reverse.go
```diff
@@ -0,0 +1,154 @@
...
+				Optional: true,
```
yanndegat: shouldn't be optional, shoudl it?

ovh/resource_ovh_ip_reverse.go
```diff
@@ -0,0 +1,154 @@
...
+			},
+			"ipreverse": {
+				Type:     schema.TypeString,
+				Required: true,
```
yanndegat: should be optional and default to the IP if ip is /32
otherwise raise an error.
it would better reflect the API

ovh/resource_ovh_ip_reverse.go
holyhope: Since ovh terraform provider already support dns resources, we should wait the apply of dns before be able to set the reverse dns. He handles retries in order to wait for dns propagation.
Am I right @pkuhner?

yanndegat: well that's not the way you should design your plan.
dns propagation is not garantueed, even if the task status is done. it could last several hours

yanndegat:
hi @pkuhner 

this looks great to me.

could you just squash your commits before we merge it pls ?

pkuhner:
> hi @pkuhner
> 
> this looks great to me.
> 
> could you just squash your commits before we merge it pls ?

Done

PR #56, opened 2018-09-25
Use zone in ID to get import working and write doc

gaetanfl:
Previsous PR was not working due to the fact that id without zone do not define a record. This PR works this around by importing using OVH_ID.zone format as explained in the updated documentation

### DIFF
--- a/ovh/resource_ovh_domain_zone_record.go
+++ b/ovh/resource_ovh_domain_zone_record.go
@@ -18,7 +18,7 @@ type OvhDomainZoneRecord struct {
 	SubDomain string `json:"subDomain,omitempty"`
 }
 
-func resourceOvhDomainZoneImportState(
+func resourceOvhDomainZoneRecordImportState(
 	d *schema.ResourceData,
 	meta interface{}) ([]*schema.ResourceData, error) {
 	givenId := d.Id()
@@ -40,7 +40,7 @@ func resourceOvhDomainZoneRecord() *schema.Resource {
 		Update: resourceOvhDomainZoneRecordUpdate,
 		Delete: resourceOvhDomainZoneRecordDelete,
 		Importer: &schema.ResourceImporter{
-			State: resourceOvhDomainZoneImportState,
+			State: resourceOvhDomainZoneRecordImportState,
 		},
 
 		Schema: map[string]*schema.Schema{
--- a/ovh/resource_ovh_domain_zone_record.go
+++ b/ovh/resource_ovh_domain_zone_record.go
@@ -2,10 +2,11 @@ package ovh
 
 import (
 	"fmt"
+	"github.com/hashicorp/terraform/helper/logging"
+	"github.com/hashicorp/terraform/helper/schema"
 	"log"
 	"strconv"
-
-	"github.com/hashicorp/terraform/helper/schema"
+	"strings"
 )
 
 type OvhDomainZoneRecord struct {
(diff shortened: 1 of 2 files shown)


ovh/resource_ovh_domain_zone_record.go
```diff
@@ -17,14 +18,29 @@ type OvhDomainZoneRecord struct {
 	SubDomain string `json:"subDomain,omitempty"`
 }
 
+func resourceOvhDomainZoneImportState(
```
yanndegat: have you run a make fmt on your code?

website/docs/r/ovh_domain_zone_record.html.markdown
```diff
@@ -45,3 +45,10 @@ The following attributes are exported:
 * `fieldType` - The type of the record
 * `ttl` - The TTL of the record
 
+## Import
+
+OVH record can be imported using the `id` and the `zone`, eg:
+
+```
+$ terraform import aws_domain_zone_record.test 1234OVH_ID.zone.tld
```
yanndegat: not aws

ovh/resource_ovh_domain_zone_record.go
gaetanfl: yes, otherwise it doesn't compile. Is there anything looking strange in this function ?

website/docs/r/ovh_domain_zone_record.html.markdown
gaetanfl: Oops, sorry, fixed

ovh/resource_ovh_domain_zone_record.go
yanndegat: ok for the format

i don't understand the split logic. and the zone.tld thing

gaetanfl: The import function takes only one argument, an id. But in order to do the import terraform needs to read a resource. And to do so it needs the zone information. To do so we need to find a trick to pass the zone. using ID.zone_name was my idea. The SplitN go function split the id given to import function in two strings, id and zone_name (which may contains dots). We pass it to the ressource created by terraform import function passed to our ImportState function and returned this Resource which will be read to proceed to the import.

yanndegat: ok great.

i think i was misreading the code because you named your func OvhDomainZone
but we're importing an OvhDomainZoneRecord right?

so if i understand it correctly, the format you suggest is

[record id].[zone name] right?

ovh/resource_ovh_domain_zone_record.go
```diff
@@ -91,6 +107,9 @@ func resourceOvhDomainZoneRecordRead(d *schema.ResourceData, meta interface{}) e
 	provider := meta.(*Config)
 
 	record := OvhDomainZoneRecord{}
+	if logging.IsDebugOrHigher() {
```
yanndegat: we rarely see this kind of perf enhancer in the terraform code base. maybe we could get rid of it

gaetanfl: I put it for Debug in fact and forget to remove. The idea behind was to understand how others providers when in debug managed to send API query they were doing which is in fact nice to understand what's going on. I removed it

ovh/resource_ovh_domain_zone_record.go
gaetanfl: Yes you're right, I just updated the function name. I can change the doc part to put your wording [record id].[zone name] if you think it's clearer

yanndegat:
wonderful. thanks a lot

PR #58, opened 2018-10-01
r/ovh_iploadbalancing_tcp_frontend: new resource

yanndegat:


yanndegat:
hi @goblain 

i'm adding this new resource
wonder if you could had a glimpse at it & review it ?

one major point: the refresh mechanism which is missing the farm/farm_server resources.
i'll make another PR to fix these too

goblain:
will look in a moment, btw. I actually have a refresh that I use, but no tests around it (https://github.com/goblain/terraform-provider-ovh/blob/workinprogress/ovh/resource_ovh_iploadbalancing_refresh.go)

goblain:
I do not specially like the idea that resource update triggers immediate refresh within it's change logic. If we have more then one change, it will add extreme delays to the operation. As I hinted above, I would lean more to having a refresh resource that can be used after all the changes were implemented, to trigger a single reload of full new config. This also helps to avoid application of partial state to the OVH loadbalancer, as it's then called only after full state is configured.

yanndegat:
you're abs. right

then we'll try to work on your refresh resource
don't you mind to make a PR, i'll merge it and try to add the missing features ?

ovh/resource_ovh_iploadbalancing_tcp_frontend.go
```diff
@@ -0,0 +1,224 @@
...
+	Port          string   `json:"port"`
+	Zone          string   `json:"zone"`
+	AllowedSource []string `json:"allowedSource,omitempty"`
+	DedicatedIpFo []string `json:"dedicatedIpfo,omitempty"`
+	DefaultFarmId int      `json:"defaultFarmId,omitempty"`
+	DefaultSslId  int      `json:"defaultSslId,omitempty"`
+	Disabled      bool     `json:"disabled,omitempty"`
+	Ssl           bool     `json:"ssl,omitempty"`
+	DisplayName   string   `json:"displayName,omitempty"`
+}
```
goblain: consider using pointers for values that you might need to update/set to their "empty" value. Otherwise you risk running into situation where you ie. created frontent with `disabled: true` now you want to enable it explicity setting `disabled: false` but... `false` is the emty/zero value for bool, so with `omitempty` this change will not get reflected on the API call.

ovh/resource_ovh_iploadbalancing_tcp_frontend_test.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+resource "ovh_iploadbalancing_tcp_frontend" "testfrontend" {
+   service_name = "${data.ovh_iploadbalancing.iplb.service_name}"
+   display_name = "%s"
+   zone = "all"
+   port = "80,443"
+   default_farm_id = "${ovh_iploadbalancing_tcp_farm.farm.id}"
+}
+`
```
goblain: as mentioned for struct, the same applies here. Consider testing setting some bool value to true and then updating it to false and verifying if it correctly applied on the ovh api.

ovh/resource_ovh_iploadbalancing_tcp_frontend.go
```diff
@@ -0,0 +1,224 @@
...
+				ForceNew: false,
+			},
+			"disabled": &schema.Schema{
+				Type:     schema.TypeBool,
+				Computed: true,
+			},
+			"ssl": &schema.Schema{
+				Type:     schema.TypeBool,
+				Computed: true,
+			},
```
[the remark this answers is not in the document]
yanndegat: youre right, done

ovh/resource_ovh_iploadbalancing_tcp_frontend_test.go
yanndegat: done

yanndegat:
thanks a lot for your review

PR #60, opened 2018-10-01
adds loadbalancer refresh

goblain:
fixes https://github.com/terraform-providers/terraform-provider-ovh/issues/59


goblain:
@yanndegat

yanndegat:
i think we can merge it as is and later see how we add tests on it
because it's now that simple to figure out


it's a kind of special resource, such as null_resource in the null provider.

PR #83, opened 2019-03-19
Added HTTP farm, servers and frontend resources

Hugoch:
Add resources to create HTTP farms, frontends and farm servers in OVH ip loadbalancer. Only TCP farms were implemented.


yanndegat:
testacc passed, merging

thanks a lot hugo

PR #92, opened 2019-10-28
update to tf 0.12.12

remijouannet:
this PR update the deps to tf 0.12.12, the last stable version of this plugin isn't compatible with the last version of terraform

i didn't "go mod vendor" before pushing


yanndegat:
thanks a lot remi

PR #93, opened 2019-11-07
Add SSH Key support

pgaxatte:
Add the data sources and resources for the /me/sshKey calls.


yanndegat:
testacc passed

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/terraform-providers/terraform-provider-ovh
 require (
 	github.com/gopherjs/gopherjs v0.0.0-20181103185306-d547d1d9531e // indirect
 	github.com/hashicorp/go-cleanhttp v0.5.1
+	github.com/hashicorp/terraform v0.12.15
 	github.com/hashicorp/terraform-plugin-sdk v1.0.0
```
yanndegat: to be removed

ovh/data_source_ovh_me_ssh_key_test.go
```diff
@@ -0,0 +1,37 @@
...
+	resource.Test(t, resource.TestCase{
+		PreCheck:  func() { testAccPreCheckCredentials(t) },
+		Providers: testAccProviders,
+		Steps: []resource.TestStep{
+			{
+				Config: fmt.Sprintf(testAccMeSshKeyDatasourceConfig, sshKey),
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(
+						"data.ovh_me_ssh_key.key_1", "key_name", sshKey),
+				),
```
yanndegat: i think that it would be better if we'd check the value of the key instead of the keyname

go.mod
pgaxatte: Done

ovh/data_source_ovh_me_ssh_key_test.go
pgaxatte: Done

website/ovh.erb
```diff
@@ -119,6 +125,15 @@
             </ul>
         </li>
 
+        <li<%= sidebar_current("docs-ovh-resource-me") %>>
+          <a href="#">Me Resources</a>
+          <ul class="nav nav-visible">
+              <li<%= sidebar_current("docs-ovh-resource-me-ssh_key") %>>
```
yanndegat: underscore-> -

website/docs/r/me_ssh_key.html.markdown
```diff
@@ -0,0 +1,38 @@
...
+
+* `key_name` - (Required) The friendly name of this SSH key.
+
+* `key` - (Required) The content of the public key in the form "ssh-algo content", e.g. "ssh-ed25519 AAAAC3...".
+
+* `default` - True when this public SSH key is used for rescue mode and reinstallations.
+
+## Attributes Reference
+
+The following attributes are exported:
```
yanndegat: usually you only reference here the additional attributes

website/ovh.erb
pgaxatte: Fixed

website/docs/r/me_ssh_key.html.markdown
pgaxatte: Hmmm I can't find a single example in the current doc where this is not written this way.

Should we leave it as is?

yanndegat:
LGTM

Thanks a lot @pgaxatte 
this is huge work

PR #94, opened 2019-11-08
Split PreCheck for acceptance tests

pgaxatte:
In order to more easily run tests selectively, I split all the pre check
tests. This way you don't need all the OVH_* env variables when testing
only a part of the provider.


yanndegat:
testacc passed on TC

PR #109, opened 2019-11-25
handle record id bigger than 32bits

kerwanp:
The OvhDomainZoneRecordImportState can have an Id bigger than 32 bits.
I did this temporary fix for my own use, and created the pull request to open the conversation.

I'm not a Go developer and just made this to continue working.


yanndegat:
hi @kerwanp 

thanks a lot for this fix. i ran the testacc and it's a green. 
i think we may have the issue elsewhere so i'll log an issue to check that all API mapped "long" attributes are correctly mapped with int64 go types

PR #113, opened 2019-12-10
r/cloud_network_private/subnet: fix acctest & rework

yanndegat:
- fix acctest: since new global regions are returned
  by the ovh_cloud_regions datasource, filter regions
  with network service UP
- use OVH_ATTACH_VRACK=0 env var to disable vrack_attach
  (useful when running tests against a project already
  attachted to a project)
- rework test with more terraform-sdk test helper funcs


yanndegat:
testacc fixed

PR #116, opened 2019-12-13
Feat/r/dedicated server update, dedicated server reboot_task

yanndegat:
add resources 

- dedicated server update: update dedicated server properties (eg: bootId)
- dedicated server reboot_task: create a reboot task


yanndegat:
testacc passed

PR #133, opened 2020-01-28
add cookie to http farm stickiness

DamienJ-wiza:
Control value for stickiness on the ressource ovh_iploadbalancing_http_farm doesn't allow to use cookie.


yanndegat:
hi @DamienJ-wiza 

thanks a lot for this contribution
test acc are green, merging

PR #151, opened 2020-08-24
Fix/cloud user

yanndegat:
- deprecate project_id in favor of service_name
- add roles attributes


yanndegat:
tests green on teamcity

PR #158, opened 2020-09-18
r/me_ssh_key: fix setting key default property and handle key not found error

invidian:


pgaxatte (changes requested):
Hello @invidian,

Thank you for your PR but I'm not sure I understand what you are trying to solve.

It seems that you want terraform to automatically re-create a key that has been manually deleted between two runs of terraform.

This is not an expected behavior of terraform: if a resource has been deleted, the user should see that it needs to be re-created and must agree on the plan to add it. There must not be any automatic creation.

Would you consider removing this part in `resourceMeSshKeyRead` but keep the rest of the fix?

invidian:
Hey @pgaxatte, thanks for the reply.

No resources will be "automatically" re-created with this PR. This is what confirmation in `terraform apply` and `terraform plan` sub-commands are for.

Most of Terraform providers handle resource life-cycle this way. If resource has been removed by the external process, Terraform will remove it from internal state and then `terraform apply` will show, that it should be created from scratch (not updated or re-created).

Currently, if one removes SSH key by hand, one must run `terraform state rm` manually to be able to run `terraform apply`, which I don't think is a correct behavior, hence the PR.

pgaxatte:
Thanks for the clarification.

You're right about the removal in the state so returning `nil` when the SSH keypair is not present is the right thing to do.

However, it would be best if you used the error code from the `GET` ([here](https://github.com/ovh/terraform-provider-ovh/blob/2f3be9d0999687a39b23e6861afdee7d1c124657/ovh/resource_ovh_me_ssh_key.go#L53)) instead of listing the keys to see if it's not present.

An example of such an error handling can be found here: https://github.com/ovh/terraform-provider-ovh/blob/2f3be9d0999687a39b23e6861afdee7d1c124657/ovh/resource_ovh_vrack_cloudproject_test.go#L62

invidian:
Right, thanks for the suggestion. I implemented it.

Perhaps similar fix should be applied to all other resources, for consistency.

pgaxatte:
You're right we'll have to sieve through all our code to see where we can apply the same reasoning

pgaxatte (approved):

PR #159, opened 2020-09-19
Add Enterprise Cloud DB Security Group And Rules

XciD:
Added support for Enterprise Cloud DB

- Security Group
- Security Group Rule

Caveat: Rule creation/update block the cluster and takes ~2 min to complete. As the cluster is bloqued, many rules to add could failed with timeout


pgaxatte (changes requested):
Hello @XciD 

Thanks for the huge work!

There is however a problem with the naming: all the resources/functions must match the full path of the API.

So for instance:
- `func dataSourceCloudDBEnterprise() ...` must become `func dataSourceCloudDBEnterpriseCluster() ...`
- `func resourceCloudDBEnterpriseSecurityGroup() ...` becomes `func resourceCloudDBEnterpriseClusterSecurityGroup() ...`

I know this is a pain to change all this but we need to keep a strong consistency on the naming.

Issue #161, opened 2020-10-10
ovh-cli is archived


telenieko:
Hi,

The documentation says:

> It is recommend to install [ovh-cli](https://github.com/ovh/ovh-cli) to handle and manage all your secret keys.

But [ovh/ovh-cli](https://github.com/ovh/ovh-cli) is an archived project with no commit since 2017. Please amend with instructions on how to acquire the credentials without ovh-cli (or have the provider do the authentication flow like other providers do).

thanks!

guillermo-menjivar:
I am in the predicament, @telenieko how were you able to generate credentials without using the `ovh-cli`?

yanndegat:
hi 

thank you for reporting this

we're going to spend some time on the provider in the next couple of weeks to fix this issue and clean the backlog of issues & PRs
but we first have to rebuild a proper CI system (since the switch to the terraform registry has deprecated the internal hashicorp ci for non official providers)

telenieko:
> I am in the predicament, @telenieko how were you able to generate credentials without using the `ovh-cli`?

You can do it from the Web: https://docs.ovh.com/gb/en/customer/first-steps-with-ovh-api/ 

> Requesting an authentication token from OVH
> Now that you have your keys, you can request a token from OVH to enable you to make requests to the API (...)

Then on `~/.ovh.conf` I have:

```ini
[default]
endpoint=ovh-eu

[ovh-eu]
application_key=SECRET
application_secret=SECRET
consumer_key=SECRET
```

It was some time ago so I don't remember the exact steps.

amstuta:
Closing this issue, doc has been updated since then.

PR #181, opened 2020-12-22
Improve readme (generation of keys)

jmMeessen:
Based on my difficulties to use the OVH provider I proposed some improvements to the readme on Terraform registry. It solves issue #161 and #9.

I describe how to retrieve the API keys and re-ordered the description about its use in the configuration.

Note that the english speaking documentation (the only one I found via Google) do barely help. The French speaking one is much better.

I wonder if there shouldn't be a hint to use the `Use Provider` button to generate the `required providers` block. Newcomers might not be familiar with the use non-official providers.

### DIFF
--- a/website/docs/index.html.markdown
+++ b/website/docs/index.html.markdown
@@ -8,32 +8,21 @@ description: |-
 
 # OVH Provider
 
-The OVH provider is used to interact with the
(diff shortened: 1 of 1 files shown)


yanndegat:
Hi @jmMeessen 

thank you very much for your contribution

it will improve the usability of the provider. thanks

PR #218, opened 2021-09-27
fix: use the right json annotation

zimbatm:
Fixes these types of warnings:

    struct field tag `json:createdAt` not compatible with reflect.StructTag.Get: bad syntax for struct tag value


yanndegat:
hi @zimbatm 

thank you very much for having fixed this.

PR #250, opened 2022-04-08
Add feature to set catalogName on order_cart

wiseflat:
I need to set a specific catalogName to order some resources.
Requirement for that is to assign a shopping cart to a loggedin client


scraly (changes requested):
Thanks :) 

I have just one remark, is it possible to change in the datasource documentation:

```
description    = "..."
```

to:
```
description    = "my cart"
```

wiseflat:
Done 👍

scraly (approved):

scraly:
Thanks, it's now ok for me :)

PR #257, opened 2022-05-09
fix: import kube cluster

### LINKED ISSUE #255: import existing ovh_cloud_project_kube fails
Importing an existing kubernetes cluster does not work due to a bug when parsing the ID.

### Affected Resource(s)
- ovh_cloud_project_kube

### Debug Output
```
Error: Import Id is not service_name/kubeid formatted
```

### Expected Behavior
Import should succeed

### Actual Behavior
Import fails

I think the root cause is a wrong check after parsing the ID to import here:
https://github.com/ovh/terraform-provider-ovh/blob/4682c886f07527f710bef4690cfb26a51365cb07/ovh/resource_cloud_project_kube.go#L99

It should check for len(splitId) to equal 2, not 3.

scraly:
Fix #255 

How to test it after fixed:

```
$ terraform import ovh_cloud_project_kube.my_kube_cluster <SERVICE_NAME>/<KUBE_CLUSTER_ID>
```

### DIFF
--- a/ovh/resource_cloud_project_kube.go
+++ b/ovh/resource_cloud_project_kube.go
@@ -96,7 +96,7 @@ func resourceCloudProjectKube() *schema.Resource {
 func resourceCloudProjectKubeImportState(d *schema.ResourceData, meta interface{}) ([]*schema.ResourceData, error) {
 	givenId := d.Id()
 	splitId := strings.SplitN(givenId, "/", 2)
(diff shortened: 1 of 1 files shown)


ovh-cds:
CDS Report terraform-provider-ovh-testacc#781.0 ✘  
* 
  * checks ✘

PR #282, opened 2022-08-08
fix containerregistry plan update pending status

thcdrt:
Hello,

I made some tests on containerregistry terraform module and it works very well !
I just found a small typo in the pending status that triggers error when upgrading plan.

Thomas


scraly:
Perfect, thanks Thomas! :-) 

With your fix, updating a container registry with terraform it's completely working without any error message.

Issue #281, opened 2022-08-04
regions_attributes of ovh_cloud_project_network_private does not contain all regions


fl42:
Hello,

### Terraform Version
```
Terraform v1.2.6
on linux_amd64
+ provider registry.terraform.io/ovh/ovh v0.19.1
```

### Affected Resource(s)
- ovh_cloud_project_network_private

### Terraform Configuration Files
```hcl
resource "ovh_cloud_project_network_private" "vnet" {
   service_name = var.service_name
   name         = "test"
   vlan_id      = 2
   regions      = ["GRA7", "SBG5"]
}

resource "ovh_cloud_project_network_private_subnet" "vnet_subnet_gra" {
   service_name = var.service_name
   network_id   = ovh_cloud_project_network_private.vnet.id
   region       = "GRA7"
   start        = "10.0.0.2"
   end          = "10.0.0.100"
   network      = "10.0.0.0/24"
   dhcp         = false
   no_gateway   = false
}

resource "ovh_cloud_project_network_private_subnet" "vnet_subnet_sbg" {
   service_name = var.service_name
   network_id   = ovh_cloud_project_network_private.vnet.id
   region       = "SBG5"
   start        = "10.0.1.2"
   end          = "10.0.1.100"
   network      = "10.0.1.0/24"
   dhcp         = false
   no_gateway   = false
}
```

```hcl
output "regions_attributes" {
  value = ovh_cloud_project_network_private.vnet.regions_attributes
}
```

### Debug Output
```sh
ovh_cloud_project_network_private.vnet: Refreshing state... [id=pn-xxx]
2022-08-04T09:43:08.632+0200 [INFO]  provider.terraform-provider-ovh_v0.19.1: 2022/08/04 09:43:08 [DEBUG] Will read public cloud private network for project: xxx, id: pn-xxx: timestamp=2022-08-04T09:43:08.632+0200
2022-08-04T09:43:08.632+0200 [INFO]  provider.terraform-provider-ovh_v0.19.1: 2022/08/04 09:43:08 [DEBUG] OVH API Request Details:
---[ REQUEST ]---------------------------------------
GET /1.0/cloud/project/xxx/network/private/pn-xxx HTTP/1.1
[...]


-----------------------------------------------------: timestamp=2022-08-04T09:43:08.632+0200
[...]

PR #286, opened 2022-08-11
fix: add hashing function on region attributes

MircoT:
With this fix is resolved the #281

### DIFF
--- a/ovh/helpers/helpers.go
+++ b/ovh/helpers/helpers.go
@@ -358,21 +358,3 @@ func ValidateSubsidiary(v string) error {
 		"ws",
 	})
 }
-
-// HashMapRegionAttributes creates an hash for the region attributes.
-func HashMapRegionAttributes(v interface{}) int {
-	attributes := v.(map[string]interface{})
-	builder := strings.Builder{}
-
-	for _, key := range []string{"status", "region", "openstackid"} {
-		value, inMap := attributes[key]
-		if inMap {
-			stringValue, ok := value.(string)
-			if ok {
-				builder.WriteString(stringValue)
-			}
-		}
-	}
-
-	return schema.HashString(builder.String())
-}
--- a/ovh/helpers/helpers.go
+++ b/ovh/helpers/helpers.go
@@ -361,11 +361,7 @@ func ValidateSubsidiary(v string) error {
 
 // HashMapRegionAttributes creates an hash for the region attributes.
 func HashMapRegionAttributes(v interface{}) int {
-	attributes, ok := v.(map[string]interface{})
-	if !ok {
-		return 0
-	}
-
+	attributes := v.(map[string]interface{})
 	builder := strings.Builder{}
 
 	for _, key := range []string{"status", "region", "openstackid"} {
--- a/ovh/helpers/helpers.go
+++ b/ovh/helpers/helpers.go
@@ -358,3 +358,25 @@ func ValidateSubsidiary(v string) error {
 		"ws",
 	})
 }
+
+// HashMapRegionAttributes creates an hash for the region attributes.
+func HashMapRegionAttributes(v interface{}) int {
+	attributes, ok := v.(map[string]interface{})
+	if !ok {
+		return 0
+	}
+
+	builder := strings.Builder{}
+
+	for _, key := range []string{"status", "region", "openstackid"} {
+		value, inMap := attributes[key]
+		if inMap {
+			stringValue, ok := value.(string)
+			if ok {
+				builder.WriteString(stringValue)
+			}
+		}
+	}
+
+	return schema.HashString(builder.String())
+}
--- a/ovh/resource_cloud_project_network_private.go
+++ b/ovh/resource_cloud_project_network_private.go
@@ -1,6 +1,7 @@
 package ovh
 
 import (
+	"errors"
 	"fmt"
 	"log"
 	"strings"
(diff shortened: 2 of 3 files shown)


scraly:
Tested and now we have all the regions displayed in `regions_attributes` field.

Before:

```
# ovh_cloud_project_network_private.vnet:
resource "ovh_cloud_project_network_private" "vnet" {
    id                 = "pn-1083678_5"
    name               = "test"
    regions            = [
        "GRA7",
        "SBG5",
    ]
    regions_attributes = [
        {
            openstackid = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxx"
            region      = "SBG5"
            status      = "ACTIVE"
        },
    ]
    regions_status     = [
        {
            region = "SBG5"
            status = "ACTIVE"
        },
    ]
    service_name       = "xxxxxxxxx"
    status             = "ACTIVE"
    type               = "private"
    vlan_id            = 5
}
```

After:

```
# ovh_cloud_project_network_private.vnet:
resource "ovh_cloud_project_network_private" "vnet" {
    id                 = "pn-1083678_5"
    name               = "test"
    regions            = [
        "GRA7",
        "SBG5",
    ]
    regions_attributes = [
        {
            openstackid = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxx"
            region      = "GRA7"
            status      = "ACTIVE"
        },
        {
            openstackid = "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxx"
            region      = "SBG5"
            status      = "ACTIVE"
        },
    ]
    regions_status     = [
        {
            region = "SBG5"
            status = "ACTIVE"
        },
    ]
    service_name       = "xxxxxx"
    status             = "ACTIVE"
    type               = "private"
    vlan_id            = 5
}
```

ovh/helpers/helpers.go
```diff
@@ -358,3 +358,25 @@ func ValidateSubsidiary(v string) error {
 		"ws",
 	})
 }
+
+// HashMapRegionAttributes creates an hash for the region attributes.
+func HashMapRegionAttributes(v interface{}) int {
+	attributes, ok := v.(map[string]interface{})
+	if !ok {
+		return 0
```
scraly, line 366: Why are you returning 0?
In which case should we obtain ok == false?
Is it possible to return an error message or something?
Thanks

scraly (changes requested):

ovh/helpers/helpers.go
MircoT: Actually, I'm returing 0 to be conformed to the [hashing function signature](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#SchemaSetFunc), that is `func(interface{}) [int](https://pkg.go.dev/builtin#int)`. Said that, the conversion check regarding the `v interface` is just to ensure that everything goes well but, in case of any failure, the only returning option for this signature is an integer and not an error. Maybe, a `-1` is a better option in place of returning `0`. 

However, since the schema should not change that conversion check is just a stretch. Let me know the way is better to modify. Also, have a code like this is an option:

```golang
// HashMapRegionAttributes creates an hash for the region attributes.
func HashMapRegionAttributes(v interface{}) int {
	attributes := v.(map[string]interface{})
	builder := strings.Builder{}

	for _, key := range []string{"status", "region", "openstackid"} {
		value, inMap := attributes[key]
		if inMap {
			stringValue, ok := value.(string)
			if ok {
				builder.WriteString(stringValue)
			}
		}
	}

	return schema.HashString(builder.String())
}
```

MircoT: Currently, I opted for that last option. I hope it is the best way to approach the problem.

scraly:
The error must be handled.

For example, the method can return an int and an error:

```
// HashMapRegionAttributes creates an hash for the region attributes.
func HashMapRegionAttributes(v interface{}) (int, error) {
    attributes, ok := v.(map[string]interface{})
    if !ok {
        return -1, errors.New("bad cast...")
    }

    builder := strings.Builder{}
...
```

MircoT:
@scraly it is not possible to handle the error in that function because in a [Terraform Schema](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#Schema) the `Set` field is a `SchemaSetFunc` that should be this [specific type](https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema#SchemaSetFunc):

```golang
type SchemaSetFunc func(interface{}) int
```

Consequently, returning a tuple instead of an integer is not an option at this level.

MircoT:
> The error must be handled.
> 
> For example, the method can return an int and an error:
> 
> ```
> // HashMapRegionAttributes creates an hash for the region attributes.
> func HashMapRegionAttributes(v interface{}) (int, error) {
>     attributes, ok := v.(map[string]interface{})
>     if !ok {
>         return -1, errors.New("bad cast...")
>     }
> 
>     builder := strings.Builder{}
> ...
> ```

Another option that doesn't involve code changes in resource creation could be something like this:

```golang
"regions_attributes": {
		Type:     schema.TypeList,
		Computed: true,
		Elem: &schema.Resource{
			Schema: map[string]*schema.Schema{
				"status": {
					Type:     schema.TypeString,
					Required: true,
				},

				"region": {
					Type:     schema.TypeString,
					Computed: true,
				},

				"openstackid": {
					Type:     schema.TypeString,
					Computed: true,
				},
			},
		},
	}
```

With such a solution, the helper function `HashMapRegionAttributes` for the set is not required and the logic behind the list creation is still the same. Otherwise, if a strict check of the element structures (beyond the schema) is necessary, the TypeMap could be used with a `SchemaValidateDiagFunc` but, in that case, the resource creation has to be changed too.

Witch one is better for this case @scraly ?

scraly:
Hi, after thinking,
is it possible to:
- move the function to `cloud_project_network_private_helpers.go` because the function should. be only used by the resource
- rename the function to `RegionAttributesHash` for example
- add an error log when the cast is failing and a panic

Thanks

scraly (changes requested):
Hi, after thinking,
is it possible to:
- move the function to `cloud_project_network_private_helpers.go` because the function should. be only used by the resource
- rename the function to `RegionAttributesHash` for example
- add an error log when the cast is failing and a panic

Thanks

MircoT:
> Hi, after thinking, is it possible to:
> 
> * move the function to `cloud_project_network_private_helpers.go` because the function should. be only used by the resource
> * rename the function to `RegionAttributesHash` for example
> * add an error log when the cast is failing and a panic
> 
> Thanks

All the requirements should be satisfied, let me know if something else is missing or if the error description is not conforming to the others in the module.

scraly:
The only thing left to comply with the whole list is to put the function in a file next to the resource, for example: `cloud_project_network_private_helpers.go` :)

scraly (changes requested):
The only thing left to comply with the whole list is to put the function in a file next to the resource, for example: cloud_project_network_private_helpers.go :)

MircoT:
> The only thing left to comply with the whole list is to put the function in a file next to the resource, for example: cloud_project_network_private_helpers.go :)

Right... I just misunderstood the previous message...

scraly (approved):
Thanks, now it's perfect :)

PR #293, opened 2022-08-26
Kube: optional update_policy parameter at cluster creation

matprig:
optional update_policy parameter at cluster creation, APIV6 was updated 

no exception thrown at cluster creation when specifying the update_policy parameter


scraly (changes requested):
Perfect, thanks :-) 

Just one typo found in the documentation (a `'` after the list of possible value)
image

Thanks

scraly (approved):
Thanks, acceptance tests are working, everything is OK :-)

PR #294, opened 2022-08-26
Add ovh_cloud_project_user_s3_policy resource

yvigara:
Add resource to manage a Public Cloud Project User's S3 policies for identity and access to S3 Object Storage resources.
See [documentation](https://docs.ovh.com/gb/en/storage/s3/identity-and-access-management/) for manual steps.


scraly:
Hi @yvigara , thanks for your PR
What do you think about adding, in this PR or in another one, a data source `ovh_cloud_project_users` (that return the list of `users_id`) and another one `ovh_cloud_project_user` in order to be able to retrieve easily users_ids and information about a user.
I think it can be helpful with this feature and s3 credentials resource? 🙂

scraly (changes requested):
Thanks for the PR :-)
With the resource, I think it can be useful to add a datasource that display policy.
With that it will be perfect :)

yvigara:
@scraly Thanks again for the feedback and the suggestions.
I have fixed the typo in the documentation and added the datasource. I'm not super pleased with the acceptance test. To fix this we would need to upgrade to a more recent version of the terrafom SDK to have access to some of the JSON validation & normalisation features. Something to consider for after the `0.20` release.

I agree with you having the `ovh_cloud_project_users` and `ovh_cloud_project_user` datasources would be great addition and make some of the example clearer. I don't think I'll be able to work on this this week. I'd rather do this in another PR.

website/docs/d/cloud_project_users.html.markdown
```diff
@@ -0,0 +1,65 @@
...
+```
+
+## Argument Reference
+
+The following arguments are supported:
+
+- `service_name` - (Required) The ID of the public cloud project. If omitted,
+  the `OVH_CLOUD_PROJECT_SERVICE` environment variable is used.
+
+## Attributes Reference
```
scraly, line 52: `openstack_id` field is displayed but empty, in my side for all users, is it normal?

website/docs/d/cloud_project_user.html.markdown
```diff
@@ -0,0 +1,50 @@
...
+---
+
+# ovh_cloud_project_user
+
+Get the user details of a previously created public cloud project user.
+
+## Example Usage
+
+```hcl
+resource "ovh_cloud_project_user" "user" {
```
scraly, line 16: You don't have to create one previously on terraform in order to use the datasource.
The goal of a datasource is originally to retrieve an existing resource in the platform :).
So this doc is for datasource so only focus to the datasource, not the resource :).

website/docs/d/cloud_project_user.html.markdown
```diff
@@ -0,0 +1,50 @@
...
+ service_name = "XXX"
+ description  = "my user"
+ role_names   = [
+  "objectstore_operator"
+ ]
+}
+
+data "ovh_cloud_project_user" "my_user" {
+ service_name = ovh_cloud_project_user.user.service_name
```
scraly, line 25:
```suggestion
 service_name = "XXX"
```

website/docs/d/cloud_project_user.html.markdown
```diff
@@ -0,0 +1,50 @@
...
+ user_id      = ovh_cloud_project_user.user.id
```
scraly, line 26: for example we can use the datasource with the list of users here in order to have the user_id or simply:

user_id      = "111111"

website/docs/d/cloud_project_user.html.markdown
```diff
@@ -0,0 +1,50 @@
...
+
+- `service_name` - (Required) The ID of the public cloud project. If omitted,
+  the `OVH_CLOUD_PROJECT_SERVICE` environment variable is used.
+
+- `user_id` - (Required) The ID of a public cloud project's user.
+
+## Attributes Reference
+
+- `creation_date` - the date the user was created.
+- `description` - See Argument Reference above.
```
scraly, line 42: id is missing?

website/docs/d/cloud_project_user_s3_policy.html.markdown
```diff
@@ -0,0 +1,59 @@
...
+  "objectstore_operator"
+ ]
+}
+
+resource "ovh_cloud_project_user_s3_credential" "my_s3_credentials" {
+ service_name = ovh_cloud_project_user.user.service_name
+ user_id      = ovh_cloud_project_user.user.id
+}
+
+resource "ovh_cloud_project_user_s3_policy" "policy" {
```
scraly, line 29: No need to create a resource in order to retrieve it through a datasource :)

website/docs/d/cloud_project_user_s3_policy.html.markdown
```diff
@@ -0,0 +1,59 @@
...
+    "Sid": "RWContainer",
+    "Effect": "Allow",
+    "Action":["s3:GetObject", "s3:PutObject", "s3:DeleteObject", "s3:ListBucket", "s3:ListMultipartUploadParts", "s3:ListBucketMultipartUploads", "s3:AbortMultipartUpload", "s3:GetBucketLocation"],
+    "Resource":["arn:aws:s3:::hp-bucket", "arn:aws:s3:::hp-bucket/*"]
+  }]
+ })
+}
+
+data "ovh_cloud_project_user_s3_policy" "policy" {
+ service_name = ovh_cloud_project_user.user.service_name
```
scraly, line 43:
```suggestion
 service_name = "XXX"
```

website/docs/d/cloud_project_user_s3_policy.html.markdown
```diff
@@ -0,0 +1,59 @@
...
+ user_id      = ovh_cloud_project_user.user.id
```
scraly, line 44:
```suggestion
 user_id      = "111111"
```

website/docs/d/cloud_project_user_s3_policy.html.markdown
```diff
@@ -0,0 +1,59 @@
...
+---
+
+# ovh_cloud_project_user_s3_policy
+
+Get the S3 Policy of a public cloud project user. The policy can be set by using the `ovh_cloud_project_user_s3_policy` resource.
+
+## Example Usage
+
+```hcl
+resource "ovh_cloud_project_user" "user" {
```
scraly, line 16: No need it.

scraly (changes requested):

scraly:
Thanks for the new resources and datasources! :) 

Just one thing, I imagined the datasource `ovh_cloud_project_users` only retrieves a list of users_id like this datasource: https://registry.terraform.io/providers/ovh/ovh/latest/docs/data-sources/cloud_project_database_users
and if user wants the details of an user, he/she have the datasource `ovh_cloud_project_user`

What do you think about that?

scraly:
Hi,
we plan to release a new version of the provider this week, can you take a look at my suggestions and request changes?
Thanks

website/docs/d/cloud_project_users.html.markdown
yvigara: It's returned by the API but not mapped to the `CloudProjectUser` struct. Removing this field for now.

yvigara:
I've updated the documentations and use the new datasource instead of creating users.

I don't think that returning only the list of `user_id` in the `ovh_cloud_project_users` datasource will be very helpful. End users would have to query each user to then be able to select a user like in the doc examples. I get this new datasource "breaks" from the standard of the majority of the datasources but as a counter example the `ovh_cloud_project_containerregistry_users` datasource returns the list of user, id and emails. Also the API returns the user details without having to do subsequent queries.
The other solution I considered was adding an optional `filter` input attribute. Happy to refactor it that way if you think it's a better option.

scraly:
Hi @yvigara , thanks for the modifications, it's OK for me :)

website/docs/d/cloud_project_user_s3_credential.html.markdown
```diff
@@ -13,24 +13,34 @@ Use this data source to retrieve the Secret Access Key of an Access Key ID assoc
...
-   user_id       = data.ovh_cloud_project_user_s3_credentials.my_s3_credentials.user_id
-   access_key_id = data.ovh_cloud_project_user_s3_credentials.my_s3_credentials.access_key_ids[0]
+  service_name  = data.ovh_cloud_project_user_s3_credentials.my_s3_credentials.service_name
+  user_id       = data.ovh_cloud_project_user_s3_credentials.my_s3_credentials.user_id
+  access_key_id = data.ovh_cloud_project_user_s3_credentials.my_s3_credentials.access_key_ids[0]
 }
 
 output "my_access_key_id" {
-   value = ovh_cloud_project_user_s3_credential.my_s3_credential.access_key_id
+  value = ovh_cloud_project_user_s3_credential.my_s3_credential.access_key_id
```
scraly, line 38:
```suggestion
  value = data.ovh_cloud_project_user_s3_credential.my_s3_credential.access_key_id
```

website/docs/d/cloud_project_user_s3_credential.html.markdown
```diff
@@ -13,24 +13,34 @@ Use this data source to retrieve the Secret Access Key of an Access Key ID assoc
...
 }
 
 output "my_secret_access_key" {
-   value     = ovh_cloud_project_user_s3_credential.my_s3_credential.secret_access_key
-   sensitive = true
+  value     = ovh_cloud_project_user_s3_credential.my_s3_credential.secret_access_key
```
scraly, line 42:
```suggestion
  value     = data.ovh_cloud_project_user_s3_credential.my_s3_credential.secret_access_key
```

scraly (changes requested):
After having checked all the final things, I have two final remain things to fix in the documentation :)
Thanks
After that I will merge your PR
Thanks again :)

yvigara:
@scraly I committed your suggested changes.

scraly:
Thanks :)

scraly (approved):
Now it's working and it's OK for me, thanks :)
It will be released in the next release

PR #300, opened 2022-09-08
fix opesearch user datasource acc test

lpatte:


scraly (approved):
Thanks, the acceptance test is now working :-)

PR #304, opened 2022-09-12
Kube/feat/customization apiserver admissionplugins

matprig:
Add feature customization at creation & update resource cluster, the customization here is to customize a part of the api server especially admission plugins

This PR also contains a fix of an helper function to avoid to bord effect and do not transform 0 value of an int pointer to a nil


scraly (approved):
Thanks :-)

PR #306, opened 2022-09-14
fix missing doc for data kube nested object customization

matprig:
fix missing doc for data kube nested object customization


scraly (approved):
Thanks :)

PR #312, opened 2022-09-20
Better indent on resources and data sources

albundy83:
Hello,

if you think it's useful, I will do the same with data sources


albundy83:
All-in :)

scraly:
I like when everything is well aligned, so I approve 🙂

scraly (approved):

PR #318, opened 2022-09-24
fix(service_name): required to optional according to the doc

TartanLeGrand:
Correct the service name according to the doc to use the var env in CI.

**Exemple**: https://registry.terraform.io/providers/ovh/ovh/latest/docs/resources/cloud_project_kube#service_name


lpatte:
Hi

`DefaultFunc` is designed to work with `required=true`. So you can not set `service_name` and set env var `OVH_CLOUD_PROJECT_SERVICE` and that work.
The issue to change required by optional is that mean that the resource should work without `service_name` or en var `OVH_CLOUD_PROJECT_SERVICE` set and that not the case.

I think it's more a doc issue

scraly:
I agree, the `service_name` (in attribute or in environment variable) is required, it's not optional.

PR #319, opened 2022-09-29
fix harshicorp go-version library usage && acceptance test

matprig:
the library harshicorp go-version did not change of version but return 3 segments even if there is only 2


scraly:
Thanks, it's perfect for me 🙂.

PR #322, opened 2022-10-05
dedicated_server: update bootid before reboot on destroy

Arnoways:
Hi,

We implement the bootid_on_destroy when we create the dedicated_server installation task but we do not use it when destroying the resource.
This pr aims to fix that.

Signed-off-by: Arnaud SINAYS <sinaysarnaud@gmail.com>


scraly (approved):
Legit :)

scraly:
Juts one thing, please add in the documentation (`index.html.markdown`) the usage of `OVH_DEDICATED_SERVER` environnement variable and what kind of information should the user set into it.

thanks

scraly:
If you plan to update the documentation, it's perfect, thanks.

PR #323, opened 2022-10-07
refactor: remove ovh_vrack_dedicated_server

Arnoways:
From the testing I did, using "ovh_vrack_dedicated_server" through terraform did not work. I tried using the API calls directly without sucess, hence this commit:

Removing this resource as the API call it executes is tagged as "LEGACY" in ovh's API and does not work anymore.
We should favor `ovh_vrack_dedicated_server_interface` which calls functional routes and avoids confusion.
I also updated the example for this resource to be more dynamic so the user will know they won't have to manually search for their interfaces' IDs.

Signed-off-by: Arnaud SINAYS <sinaysarnaud@gmail.com>


scraly (changes requested):

scraly (approved):

scraly:
Thanks :)

PR #328, opened 2022-10-18
replace resizing status to upscaling/downscaling

matprig:
the 25th of october, a change will be prodded to replace node pool RESIZING status to DOWNSCALING and UPSCALING status.

pending state change has to be adapt

NOTA: when running test do not forget to raise go test timeout with "-timeout 30m"


scraly:
Thanks :)

scraly (approved):

PR #338, opened 2022-11-09
feat(go-ovh): add User-Agent on go-ovh client to identify calls that comes from Terraform

rbeuque74:
Since most of Terraform providers are customising the User-Agent to detect the calls that comes from Terraform usage, we should do the same on the terraform-provider-ovh.

It will include the Terraform provider version, that will ease the debugging purpose when customers comes to us with problems.


yanndegat:
thanks a lot @rbeuque74

PR #356, opened 2023-01-04
Add support for HTTP header on resource_iploadbalancing_http_frontend

wilfriedroset:
This commit adds support for `httpHeader` attribute when manipulating http frontend.

See: https://docs.ovh.com/fr/load-balancer/http-headers

### DIFF
--- a/website/docs/r/iploadbalancing_http_frontend.html.markdown
+++ b/website/docs/r/iploadbalancing_http_frontend.html.markdown
@@ -34,6 +34,31 @@ resource "ovh_iploadbalancing_http_frontend" "testfrontend" {
 }
 ```
 
+## Example Usage with HTTP header
+
+```
+data "ovh_iploadbalancing" "lb" {
+  service_name = "ip-1.2.3.4"
+  state        = "ok"
+}
+
+resource "ovh_iploadbalancing_http_farm" "farm80" {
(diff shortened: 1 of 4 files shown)


website/docs/r/iploadbalancing_http_frontend.html.markdown
```diff
@@ -51,6 +51,7 @@ The following arguments are supported:
 * `disabled` - Disable your frontend. Default: 'false'
 * `ssl` - SSL deciphering. Default: 'false'
 * `redirect_location` - Redirection HTTP'
+* `http_header` - HTTP headers to add to the frontend. List of string.
```
scraly, line 54: Is it possible to have an example with an http_header in this documentation page? :-)

scraly (changes requested):

website/docs/r/iploadbalancing_http_frontend.html.markdown
```diff
@@ -51,6 +51,7 @@ The following arguments are supported:
...
 
 ## Attributes Reference
```
scraly, line 56: Moreover, is it possible to add missing attributes in the list please.

`port`, `service_name` and `zone` seems to be missing.

scraly (changes requested):

scraly (approved):
Perfect :)

PR #357, opened 2023-01-06
fix 3 issue for nodepool deserialization

matprig:
fix error text
fix same schema resource/data
fix deserialization


scraly (approved):
Top!

Issue #335, opened 2022-11-09
Support for specifying database "advanced configuration"


d0x2f:
Hello,

As a feature request, I'd like the ability to specify the same "advanced configuration" options available in the web interface via terraform.

For me specifically, I need to configure the sql_mode & sql_require_primary_key settings of a mysql database.
I'm looking into doing this with a script via the rest api for now.

Thanks!

### Terraform Version
$ terraform -v
Terraform v1.3.4
on linux_amd64

registry.terraform.io/ovh/ovh 0.22.0

lpatte:
Hello,

Thanks for the request, that a MUST for sure. 
The request is now tracked on our side and should be implement for a future release.

yomovh:
Shipped in 0.28.0, closing the issue

biapar:
Don't work. I've MYSQL Public cloud and I've this sql mode:
{
  "mysql.sql_mode": "ANSI,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION,NO_ZERO_DATE,NO_ZERO_IN_DATE,STRICT_ALL_TABLES",
  "mysql.sql_require_primary_key": "false"
}
but in MYSQL there is:
'REAL_AS_FLOAT,PIPES_AS_CONCAT,ANSI_QUOTES,IGNORE_SPACE,ONLY_FULL_GROUP_BY,ANSI,STRICT_ALL_TABLES,NO_ZERO_IN_DATE,NO_ZERO_DATE,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION'

PR #360, opened 2023-01-12
add advanced config and manage avnadmin user

lpatte:
- Add Advanced Configuration (Resource/datasource/doc) (https://github.com/ovh/terraform-provider-ovh/issues/335)
- Manage avnadmin user (Resource)
- Some fix and  factorization for user resources

### DIFF
--- a/website/docs/d/cloud_project_database_users.html.markdown
+++ b/website/docs/d/cloud_project_database_users.html.markdown
@@ -31,11 +31,6 @@ output "user_ids" {
 
 * `engine` - (Required) The engine of the database cluster you want to list users. To get a full list of available engine visit:
 [public documentation](https://docs.ovh.com/gb/en/publiccloud/databases).
-Available engines:
-  * `cassandra`
-  * `kafka`
-  * `kafkaConnect`
-  * `mysql`
 
 * `cluster_id` - (Required) Cluster ID
 
--- a/website/docs/r/cloud_project_database.html.markdown
+++ b/website/docs/r/cloud_project_database.html.markdown
@@ -94,6 +94,10 @@ resource "ovh_cloud_project_database" "mysqldb" {
     region  = "SBG"
   }
   flavor        = "db1-4"
+  advanced_configuration = {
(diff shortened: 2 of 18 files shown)


website/docs/d/cloud_project_database_users.html.markdown
```diff
@@ -29,13 +29,7 @@ output "user_ids" {
 * `service_name` - (Required) The id of the public cloud project. If omitted,
   the `OVH_CLOUD_PROJECT_SERVICE` environment variable is used.
 
-* `engine` - (Required) The engine of the database cluster you want to list users. To get a full list of available engine visit:
```
scraly, line 32: Why are you removing the list of available engines? :)

website/docs/r/cloud_project_database.html.markdown
```diff
@@ -94,6 +94,10 @@ resource "ovh_cloud_project_database" "mysqldb" {
     region  = "SBG"
   }
   flavor        = "db1-4"
+  advanced_configuration = {
+    "mysql.sql_mode": "ANSI,ERROR_FOR_DIVISION_BY_ZERO,NO_ENGINE_SUBSTITUTION,NO_ZERO_DATE,NO_ZERO_IN_DATE,STRICT_ALL_TABLES",
```
scraly, line 98: What about the available advanced configuration for other engine than mysql? :)

scraly (changes requested):

scraly (changes requested):
added several questions :)

website/docs/d/cloud_project_database_users.html.markdown
lpatte: I re-add full list doc, that was a mistake to remove it). 
But remove `Available engines`list because all engines are available for this datasource.

website/docs/r/cloud_project_database.html.markdown
lpatte: There is too much advanced configuration, especially if we add up all the engines. 
Each engine already has a page for advanced parameters in the public documentation.
Here is just an example to how use it and can be adapted to all other advanced parameters

scraly (approved):

scraly:
Thanks :)

biapar:
avnadmin don't have SUPER permissions: Error Code: 1227. Access denied; you need (at least one of) the SUPER or SYSTEM_VARIABLES_ADMIN privilege(s) for this operation

PR #372, opened 2023-01-31
typo: missing prefix

ldicarlo:
So I am not an expert in terraform but I think you are missing the prefix here.


scraly (approved):
Thanks, that's true :-)

PR #373, opened 2023-02-03
feat(kube): Let user configure kube proxy

SkYNewZ:
- Added a new argument in resource `ovh_cloud_project_kube` called `customization_kube_proxy` to customize whether to use `iptabes` or `ipvs` and related settings
- Add a new argument `kube_proxy_mode` to choose between `iptables` or `ipvs`
- TypeSet `customization.apiserver` is now `customization_apiserver`

### DIFF
--- a/ovh/data_cloud_project_kube.go
+++ b/ovh/data_cloud_project_kube.go
@@ -92,8 +92,7 @@ func dataSourceCloudProjectKube() *schema.Resource {
 							Optional: true,
 							ForceNew: false,
 							MaxItems: 1,
-							Set:      CustomIPtablesSchemaSetFunc(false),
-
+							Set:      CustomIPVSIPTablesSchemaSetFunc(),
 							Elem: &schema.Resource{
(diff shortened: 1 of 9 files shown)


scraly (changes requested):
Hi and thank you :) 
Is it possible to add in resource's documentation a special example to show how to create a kubernetes cluster with this new feature?

Like this one:
image

Thanks :)

scraly:
Changing the name of a field is not painless because users already use them in many and many of terraform configuration files.
I understand now we have two possible customization:  `customization_kube_proxy`  and `customization_apiserver ` and possibly in the future another customization possibilities... but is it possible to not rename the existing one (for existing users)?

Thanks

scraly (changes requested):
according to our discussion, a proposition can be to add a deprecated field and add the new one as an hybride mode
thanks! :)

scraly (approved):
perfect :)

scraly:
reverted - need some fix for update case

PR #374, opened 2023-02-07
Use user-provided timeouts for all resources

### LINKED ISSUE #332: terraform reached the timeout when creating the kube cluster
Hello,

I'm trying to create a managed cluster using this provider but the creation is very long and reched the default 20min timeout.
I've try to set the timeouts block on node pool and kube resource without success. The providr is not supporting this block for that resources.

```
  timeouts {
    create = "60m"
    update = "60m"
    delete = "60m"
  }
```

Is there a way to solve this ?

SkYNewZ:
Fix #332

### DIFF
--- a/ovh/resource_cloud_project_kube.go
+++ b/ovh/resource_cloud_project_kube.go
@@ -33,6 +33,14 @@ func resourceCloudProjectKube() *schema.Resource {
 			State: resourceCloudProjectKubeImportState,
 		},
 
+		Timeouts: &schema.ResourceTimeout{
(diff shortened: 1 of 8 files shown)


scraly (changes requested):
Thanks for the PR :) 
Is it possible also to add the new possibility of timeout in the documentation page (for kube, iprestrictions, nodepool and oidc)?

Example:
https://registry.terraform.io/providers/ovh/ovh/latest/docs/resources/cloud_project_database#timeouts

Thanks

scraly (approved):

PR #396, opened 2023-03-09
fix(docs) ovh_domain_zone_record: remove quote for ttl value in example

Arnoways:
It should be an int but example can be misleading.

# Description

Updating example inside documentation for ttl value in ovh_domain_zone_record.

## Type of change

- Documentation update


scraly (approved):

scraly:
Indeed the TTL should not be a string :) 

```
resource "ovh_domain_zone_record" "test" {
    fieldtype = "A"
    subdomain = "test"
    target    = "0.0.0.0"
    ttl       = 3600
    zone      = "test.ovh"
}
```

PR #430, opened 2023-06-09
fixes bug in TestAcc for iploadbalancing http/tcp server

Arnoways:
Removing the introduced test in #428, it turns out the whole API response is written inside the resource. Including the null values. Which means the TestCheckNoResourceAttr was not fitting. Not replacing it as there's no built-in func that allows for checking that a resource is set while having a nil value.


scraly (approved):
OK :)

PR #434, opened 2023-06-26
Implements interactions with BringYourOwnImage endpoint

Eclion:
# Description

This PR implements the interactions with the `/dedicated/server/${server_ns}/bringYourOwnImage` endpoint.
The goal of such implementation is to be able to install custom images on servers.

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] Documentation update

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x] Test A: `make testacc TESTARGS="-run TestAccDedicatedServerBringYourOwnImage"`
image

**Test Configuration**:
* Terraform version: `terraform version`: local build following the README.md
```hcl
# Configure the OVHcloud Provider
terraform {
  required_providers {
    ovh = {
      source  = "terraform.local/local/ovh"
      version = "0.0.1"
    }
  }
}

provider "ovh" {
  endpoint           = "ovh-eu"
  application_key    = "TO_FILL:app_key"
  application_secret = "TO_FILL:app_secret"
  consumer_key       = "TO_FILL:consumer_key"
}

variable "url" {
  description = "img url"
  type        = string
  sensitive   = false
  default = "TO_FILL:img_url"
}

data "ovh_dedicated_server" "server" {
  service_name = "TO_FILL:server_ns"
}

resource ovh_dedicated_server_bringyourownimage "server_install" {
  service_name = data.ovh_dedicated_server.server.name
  url = var.url
  type = "qcow2"
  config_drive {
    enable = true
    hostname  = "TO_FILL:server_name"
    ssh_key   = file("TO_FILL:server_ssh_key")
    user_data = "TO_FILL:custom_data"
    user_metadatas = {
      TO_FILL = "custom_metadata"
    }
  }
}

output "out" {
  value = resource.ovh_dedicated_server_bringyourownimage.server_install
  sensitive = true
}

```

# Checklist:


ovh/resource_dedicated_server_bringyourownimage_test.go
```diff
@@ -0,0 +1,63 @@
...
+						"ovh_dedicated_server_bringyourownimage.os_install", "status", "done"),
+				),
+			},
+		},
+	})
+}
+
+func testAccDedicatedServerBringYourOwnImageConfig() string {
+	dedicated_server := os.Getenv("OVH_DEDICATED_SERVER")
+	url := os.Getenv("OVH_BYOI_IMAGE_URL")
```
scraly, line 36:
```suggestion
	url := os.Getenv("OVH_DEDICATED_SERVER_BYOI_IMAGE_URL")
```

scraly: An hardcoded variable might be enough.
If not, I think we need to keep the coherence of env variable name.

ovh/resource_dedicated_server_bringyourownimage_test.go
```diff
@@ -0,0 +1,63 @@
...
+	`,
+		dedicated_server,
+		url,
+		hostName,
+	)
+}
+
+func testAccPreCheckByoiImageUrl(t *testing.T) {
+	testAccPreCheckCredentials(t)
+	checkEnvOrSkip(t, "OVH_BYOI_IMAGE_URL")
```
scraly, line 62:
```suggestion
	checkEnvOrSkip(t, "OVH_DEDICATED_SERVER_BYOI_IMAGE_URL")
```

scraly (changes requested):
Thanks. It's working, just few remarks and suggestions :)

And a question:

When a user change the URL of the image, should not be re-created the full resource and not "modify" the URL (but do nothing on the API side)?

sbraz:
Hi @Eclion and thanks again for your contribution. I am really sorry but we cannot merge this as-is because the `bringYourOwnImage` routes will be deprecated soon (by the end of 2023).

We (the bare metal team) are working on making BYOI installable via the normal `/dedicated/server/{serviceName}/install/start` routes. It will make it easier for you as you won't need to guess the task ID any more, for instance :). I will notify you once this is possible, but, in the meantime, we do not want to add more dependencies that would make it harder to deprecate the old routes.

Eclion:
Hi @scraly,
Thanks you very much for your review!

> When a user change the URL of the image, should not be re-created the full resource and not "modify" the URL (but do nothing on the API side)?

I arbitrarily decided that when the URL (or any other field in fact) changes, the install process should be triggered.
This way, users can use `ignore_changes` in the lifecycle of their resource if they wish to ignore updates in the URL or other field.

The reason why it isn't just an update of the resource but a renew of the resource is because I use the ID of the task for as the ID of the resource, hence the resource replacement when updating the URL!

Obviously, that's a decision I arbitrarily took, and I would be more then happy to discuss alternatives :)

___

Hi @sbraz !

Understood. Too bad for the PR but I am more than happy to know we won't depend on the task ID!
(it adds a bit of complexity to get the latest install task after all :) )

I will patiently wait for future updates then!

scraly:
As we discussed with @sbraz , I will close this PR because the route will be deprecated soon.
@sbraz please create a new PR when a change in the API occurs :-).

PR #473, opened 2023-09-20
Update the ovh_dedicated_server_boots documentation

antham:
# Description

Update the ovh_dedicated_server_boots documentation to add a missing parameter `kernel` which is defined but not exposed in the documentation.

## Type of change

- [X] Documentation update

# How Has This Been Tested?

Not tested this a documentation change

# Checklist:

- [X] My code follows the style guidelines of this project
- [X] I have performed a self-review of my code
- [ ] My changes generate no new warnings or issues
- [ ] New and existing acceptance tests pass locally with my changes
- [ ] I ran succesfully `go mod vendor` if I added or modify `go.mod` file

### DIFF
--- a/website/docs/d/dedicated_server_boots.html.markdown
+++ b/website/docs/d/dedicated_server_boots.html.markdown
@@ -21,6 +21,8 @@ data "ovh_dedicated_server_boots" "netboots" {
 
 * `boot_type` - (Optional) Filter the value of bootType property (harddisk, rescue, ipxeCustomerScript, internal, network)
 
+* `kernel` - (Optional) Filter the value of kernel property (iPXE script name)
+
 ## Attributes Reference
 
 The following attributes are exported:


amstuta:
@antham the parameter `kernel` has been removed a while ago from the API, as you can see in the spec of the route here: https://api.ovh.com/console-preview/?section=%2Fdedicated%2Fserver&branch=v1#get-/dedicated/server/-serviceName-/boot
As a result, I think we should remove the parameter instead of updating the documentation

antham:
Ok I thought it was the other way around. This is a bit a problem because without this parameter you cannot point out a specific script if you have several or I missed something.

amstuta:
My bad, I misread the datasource: all the scripts are fetched and then filtered using the `kernel` parameter. I'll merge your PR then.

amstuta (approved):

PR #536, opened 2024-01-15
Fix mongodb role that constantly update at each Update

lpatte:
# Description

Fix an Issue Reported to me where some MongoDB role constantly update (ie: clusterMonitor@admin).
The reason was because we trimed "@admin" in the output because on API side we have some "builtin" role that the user can use without add the "@admin" (it is syntactic sugar).
But of course that break all other role with admin as authentication database that not a builtin role.

So to be more consistent, I remove this trim and force user to alway add the authentication database.
Like this it's easier to check update and clearer for the user.

## Type of change

- Bug fix (non-breaking change which fixes an issue)
- Breaking change (fix or feature that would cause existing functionality to not work as expected)
  - Role parameter has now a verifcationFunc, so a terraform with a builtin role without authentication Database will send an error and should be updated by adding "@admin".
- Documentation update

# How Has This Been Tested?

I locally build Terraform  and test with following roles resources to apply several time:

```hcl
# Work and do not update at each apply
resource "ovh_cloud_project_database_mongodb_user" "monitoring" {
  service_name = "XXXXXXXXX"
  cluster_id   = "XXXXXXX"
  name         = "monitoring"
  roles        = ["restore@admin", "clusterMonitor@admin", "read@local"]
}

# restore is a builtin but Terraform send a error because resource now force to add authentication Databse
resource "ovh_cloud_project_database_mongodb_user" "monitoring" {
  service_name = "XXXXXXXX"
  cluster_id   = "XXXXXXXX"
  name         = "monitoring"
  roles        = ["restore", "clusterMonitor@admin", "read@local"]
}
```


BellionBastien (dismissed):

ovh/resource_cloud_project_database_mongodb_user.go
```diff
@@ -63,10 +64,17 @@ func resourceCloudProjectDatabaseMongodbUser() *schema.Resource {
 			},
 			"roles": {
 				Type:        schema.TypeSet,
-				Description: "Roles the user belongs to (without authentication database)",
+				Description: "Roles the user belongs to with the authentication database",
 				Optional:    true,
 				Elem: &schema.Schema{
 					Type: schema.TypeString,
+					ValidateFunc: func(v interface{}, k string) (ws []string, errors []error) {
```
amstuta, line 71: instead of this error, couldn't you manually add the `@admin` part to the role when it is not present ? This way we would avoid the breaking change.

BellionBastien: I think for this one we won't be able to dodge the breaking change.
We should not imply the @admin for the customer as it determine on which database the user have rights. For this reason we would prefer customer to be fully aware of the value it will be set to.

amstuta:
seen together, the doc should be updated to inform users that there is a breaking change in the new version

amstuta:
@lpatte tests `TestAccCloudProjectDatabaseMongodb*` must be updated to add the `@db` part in the roles

amstuta (approved):

PR #537, opened 2024-01-15
fix: Stop computing URNs in the provider and rely on the API responses

amstuta:
# Description

The URNs of the IAM resources are computed locally in the provider. This is not necessary anymore since a field `iam` is included in the response bodies of the API, which includes the `urn`.
Also added a few `url.PathEscape` on service ids.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

# How Has This Been Tested?

Added a check in all tests that check resources that have an IAM URN.

# Checklist:

### DIFF
--- a/ovh/data_dedicated_ceph.go
+++ b/ovh/data_dedicated_ceph.go
@@ -88,7 +88,7 @@ func dataSourceDedicatedCephRead(d *schema.ResourceData, meta interface{}) error
 	}
 	log.Printf("[DEBUG] CEPH is %v", ceph.CephMonitors)
 	d.SetId(ceph.ServiceName)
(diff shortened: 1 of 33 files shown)


rbeuque74 (approved):

amstuta:
> We can also remove the compute URN function from the utils ;)

It's not possible right now because it is still used in datasources `ovh_dbaas_logs_cluster` and `ovh_dbaas_logs_clusters`. I'm not sure adding the call to `GET /dbaas/logs/{serviceName}` in these datasouces just to retrieve the services' URNs is a good idea, so I kept this util func for now.

PR #547, opened 2024-01-24
"service_name" could not be both required and have a default function

nervo:
# Description

In some resources such as managed databases, a `service_name`  argument is both required AND have default function. Those are not  compatible, and could lead to a shrodinger situation where a end user provide its value via an env var (`OVH_CLOUD_PROJECT_SERVICE`), and see its resource tagged as invalid.

My proposition is to remove the `required` flag.

Note that this pr only applies to `ovh_cloud_project_database` , but the same should be applied to all relevants situations.

Also, see: https://github.com/hashicorp/terraform-plugin-docs/issues/42 (thx @hyvs )

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# Checklist:

- [ ] My changes generate no new warnings or issues
- [ ] New and existing acceptance tests pass locally with my changes
- [ ] I ran succesfully `go mod vendor` if I added or modify `go.mod` file

### DIFF
--- a/ovh/resource_cloud_project_database.go
+++ b/ovh/resource_cloud_project_database.go
@@ -34,7 +34,6 @@ func resourceCloudProjectDatabase() *schema.Resource {
 			"service_name": {
 				Type:        schema.TypeString,
 				ForceNew:    true,
-				Required:    true,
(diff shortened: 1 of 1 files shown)


nervo:
Forget it, this is more a problem with vscode-terraform, see: https://github.com/hashicorp/vscode-terraform/issues/1616

PR #582, opened 2024-03-15
clean deprecated function of cloud_project_database resources

lpatte:
# Description

After running `staticcheck . | grep cloud_project_database` I got some errors

1. hashcode.String and  hashcode.Strings are deprecated.
2. error strings should not be capitalized ([ST1005](https://staticcheck.io/docs/checks#ST1005))
3. error should be returned as the last argument ([ST1008](https://staticcheck.io/docs/checks#ST1008))
4. several `resource.<func>` are deprecated and should be replace by `retry.<func>`

For `hashcode` function it look like, following deprecation comment, functions was copied in this project but without remove the deprecated comment (840501c51f52536a859c344298b3c663687f57a4), I update the comment.

## Type of change


ovh/resource_cloud_project_database_ip_restriction.go
```diff
@@ -79,7 +79,7 @@ func resourceCloudProjectDatabaseIpRestrictionImportState(d *schema.ResourceData
 	n := 4
 	splitId := strings.SplitN(givenId, "/", n)
 	if len(splitId) != n {
-		return nil, fmt.Errorf("Import Id is not service_name/engine/cluster_id/ip formatted")
+		return nil, fmt.Errorf("mport Id is not service_name/engine/cluster_id/ip formatted")
```
amstuta, line 82:
```suggestion
		return nil, fmt.Errorf("import Id is not service_name/engine/cluster_id/ip formatted")
```

amstuta:
Nice cleaning job :)

amstuta (approved):

PR #615, opened 2024-04-11
Improve error message : add request ID

theredcat:
# Description

Add request ID in error message when CheckDeleted function is used. This information is systematically asked by the support but isn't available anywhere in a normal run

## Type of change

-  Improvement (improve existing resource(s) or datasource(s))

# How Has This Been Tested?

Just waited for one of the occasionnal API fail

### DIFF
--- a/ovh/helpers/helpers.go
+++ b/ovh/helpers/helpers.go
@@ -362,9 +362,13 @@ func ConditionalAttributeBool(buff *bytes.Buffer, name string, val *bool) {
 // CheckDeleted checks the error to see if it's a 404 (Not Found) and, if so,
 // sets the resource ID to the empty string instead of throwing an error.
 func CheckDeleted(d *schema.ResourceData, err error, endpoint string) error {
-	if errOvh, ok := err.(*ovh.APIError); ok && errOvh.Code == 404 {
-		d.SetId("")
-		return nil
+	errOvh, ok := err.(*ovh.APIError)
+	if ok {
+		if errOvh.Code == 404 {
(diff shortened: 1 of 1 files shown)


ovh/helpers/helpers.go
```diff
@@ -362,12 +362,13 @@ func ConditionalAttributeBool(buff *bytes.Buffer, name string, val *bool) {
...
 func CheckDeleted(d *schema.ResourceData, err error, endpoint string) error {
-	if errOvh, ok := err.(*ovh.APIError); ok && errOvh.Code == 404 {
+	errOvh, ok := err.(*ovh.APIError);
+	if ok && errOvh.Code == 404 {
 		d.SetId("")
 		return nil
 	}
 
-	return fmt.Errorf("calling %s:\n\t %s", endpoint, err.Error())
+	return fmt.Errorf("calling %s:\n\t %s Query ID : %s", endpoint, err.Error(), errOvh.QueryID)
```
amstuta, line 371: you should use `errOvh.QueryID` only if we're sure that `ok` is true, something like:
```
var queryID string
if ok {
  queryID = errOvh.QueryID
}
```

mxpetit (changes requested):
agreed with @amstuta

theredcat:
@mxpetit @amstuta I've made some changes. I've preffered to split the error message in two cases to prevent an empty `Query ID :` when it's not available

amstuta:
Thanks for your contribution @theredcat :)

amstuta (approved):

PR #623, opened 2024-04-19
fix kafka + add kafka_schema_registry attribute

lpatte:
# Description

- Fix an issue on Kafka update and/or create (when kafka_rest_api or advanced_configuration set).
  - Kafka update API does not have `backups` field, this is a corner case of the resource `ovh_cloud_project_database`
- Add kafka_schema_registry attribute, following API

Fixe an issue reported directly to me.

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Improvement (improve existing resource(s) or datasource(s))
- [ ] Documentation update

### DIFF
--- a/ovh/types_cloud_project_database.go
+++ b/ovh/types_cloud_project_database.go
@@ -67,69 +67,67 @@ type CloudProjectDatabaseResponse struct {
 	Description           string                                      `json:"description"`
 	Endpoints             []CloudProjectDatabaseEndpoint              `json:"endpoints"`
 	Flavor                string                                      `json:"flavor"`
-	Id                    string                                      `json:"id"`
+	ID                    string                                      `json:"id"`
 	IPRestrictions        []CloudProjectDatabaseIPRestrictionResponse `json:"ipRestrictions"`
 	MaintenanceTime       string                                      `json:"maintenanceTime"`
-	NetworkId             string                                      `json:"networkId"`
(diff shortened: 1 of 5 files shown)


ovh/types_cloud_project_database.go
```diff
@@ -301,10 +305,13 @@ func (opts *CloudProjectDatabaseUpdateOpts) FromResource(d *schema.ResourceData)
 	if err != nil {
 		return nil, err
 	}
+	time := d.Get("backup_time").(string)
 
-	opts.Backups = CloudProjectDatabaseBackups{
-		Regions: regions,
-		Time:    d.Get("backup_time").(string),
+	if engine != "kafka" && (len(regions) != 0 || time != "") {
```
amstuta, line 310: not the same check as in `CloudProjectDatabaseCreateOpts.fromResource` ? (no `engine != "kafka"` there)

lpatte: `backups` field is ignored for a lot of engine in API (because the engine has not backup or backup not configurable) in POST and PUT, Kafka is the only engine where the schema of the PUT do not has "backups" fields. So this `engine != "kafka"` will ignore backups attribute for `Kafka` when update, no need to add this in CloudProjectDatabaseCreateOpts.fromResource because our API already ignore this field

amstuta (approved):

PR #672, opened 2024-06-27
fix: ovh_domain_zone_record: use correct default value for ttl

### LINKED ISSUE #666: [BUG] ovh_domain_zone_record: can't use ttl default value of '0'
### Describe the bug
When a DNS zone record is defined in terraform where the terraform TTL is identical to the default TTL for the DNS zone in OVH, duplicated records are created.
Cause is most likely, that OVH internally represents records that use the default TTL with a TTL of '0'. In the "ovh_domain_zone_record" resource, TTL is required to be > 60, and thus cannot be set to '0'.

### Terraform Version

`code`
Terraform v1.8.4
on windows_amd64
`code`

### OVH Terraform Provider Version

`code`
.\terraform.exe init

Initializing the backend...

Initializing provider plugins...
- Reusing previous version of ovh/ovh from the dependency lock file
- Using previously-installed ovh/ovh v0.45.0
`code`

### Affected Resource(s)

Please list the resources as a list, for example:
- ovh_domain_zone_record

### Terraform Configuration Files

```hcl
terraform {
  required_providers {
    ovh = {
      source = "ovh/ovh"
    }
  }
}


provider "ovh" {
  endpoint           = "ovh-eu"
  application_key    = "****"
  application_secret = "****"
  consumer_key       = "****"
}


resource "ovh_domain_zone_record" "test" {
  zone      = "<dns-zone-name>"
  subdomain = "test"
  fieldtype = "A"
  ttl       = 3600
  target    = "<some ip>"
}

```

### Debug Output

No debug info available.

### Panic Output

No panic output.

### Expected Behavior

The provider should have noticed that there already is a matching domain zone record and not have created a second entry.

### Actual Behavior

A second domain zone record has been created with a TTL of 3600, although a record already existed with a TTL of '0', which represents the default, which is '3600'.

### Steps to Reproduce

Please list the steps required to reproduce the issue, for example:
1. Create an "ovh_domain_zone_record" for a DNS entry that has been created in the web-UI and that uses the default TTL of the DNS zone
2. `terraform apply`

rbeuque74:
Fixes #666

### DIFF
--- a/ovh/resource_domain_zone_record.go
+++ b/ovh/resource_domain_zone_record.go
@@ -72,11 +72,11 @@ func resourceOvhDomainZoneRecord() *schema.Resource {
 			"ttl": {
 				Type:     schema.TypeInt,
 				Optional: true,
-				Default:  3600,
+				Default:  0,
 				ValidateFunc: func(val interface{}, key string) (warns []string, errs []error) {
 					v := val.(int)
-					if v < 60 {
-						errs = append(errs, fmt.Errorf("%q must be greater than or equal to 60, got: %d", key, v))
+					if v < 60 && v != 0 {
+						errs = append(errs, fmt.Errorf("%q must be either equal to 0 or, greater than or equal to 60, got: %d", key, v))
 					}
 					return
 				},
--- a/ovh/resource_domain_zone_record_test.go
+++ b/ovh/resource_domain_zone_record_test.go
@@ -124,6 +124,34 @@ func TestAccDomainZoneRecord_Basic(t *testing.T) {
 						"ovh_domain_zone_record.foobar", "ttl", "3600"),
 				),
 			},
+			{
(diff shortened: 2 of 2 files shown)


ovh/resource_domain_zone_record.go
```diff
@@ -72,10 +72,10 @@ func resourceOvhDomainZoneRecord() *schema.Resource {
 			"ttl": {
 				Type:     schema.TypeInt,
 				Optional: true,
-				Default:  3600,
+				Default:  0,
 				ValidateFunc: func(val interface{}, key string) (warns []string, errs []error) {
 					v := val.(int)
-					if v < 60 {
+					if v < 60 && v != 0 {
 						errs = append(errs, fmt.Errorf("%q must be greater than or equal to 60, got: %d", key, v))
```
amstuta, line 79: shouldn't error message be updated as well ?

amstuta (approved):

PR #965, opened 2025-04-30
feat: Add support for password rotation in cloud project user resource

### LINKED ISSUE #962: [FEATURE] Add support for password rotation on ovh_cloud_project_user via rotate_when_changed
### Description

Currently, the ovh_cloud_project_user resource does not support a mechanism for automatic password rotation triggered by external conditions. This is important for adhering to security best practices and automating credential lifecycle management.

Adding support for an argument similar to rotate_when_changed, as implemented in the AzureAD provider, would enable password rotation through external triggers such as a timestamp, improving automation and compliance with security standards. ([rotate_when_changed – azuread_service_principal_password](https://registry.terraform.io/providers/hashicorp/azuread/latest/docs/resources/service_principal_password#rotate_when_changed-1))

### Affected Resource(s) and/or Data Source(s)

* ovh_cloud_project_user

### Potential Terraform Configuration

```hcl
resource "ovh_cloud_project_user" "example" {
  service_name = var.project_Id
  description  = "Service User created by Terraform"

  rotate_when_changed = {
    last_rotation = var.last_rotation
  }
} 
```

A map of arbitrary key/value pairs that will force recreation of the password when they change, enabling password rotation based on external conditions such as a rotating timestamp.

### Additional context

This follows the same pattern used in other Terraform providers. For example, the AzureAD provider supports the rotate_when_changed argument for azuread_service_principal_password.

Implementing this in the ovh_cloud_project_user resource would allow users to safely automate password management through Terraform, without requiring manual deletion or external scripting. It enhances security hygiene and simplifies integration into CI/CD pipelines.

pascalinthecloud:
# Description

This PR introduces a password rotation mechanism for the ovh_cloud_project_user resource. A new optional attribute rotate_when_changed has been added. When any key/value pair within this map changes, the provider will trigger the OVHcloud API to regenerate the user's password without recreating the user resource itself.

This allows for scheduled or event-driven password rotation. The new password is stored in the Terraform state, making it available for use by other resources (e.g., storing it securely).

Fixes #962 

## Type of change
- [x] Documentation update

# How Has This Been Tested?

go test -v ./ovh -run '^TestAccCloudProjectUser_.*$' -timeout 30m

```
=== RUN   TestAccCloudProjectUser_basic
    provider_test.go:506: Read Cloud Project /cloud/project/aa2425907763475a89a4bf4e9f9e22fb -> status: 'ok', desc: 'itsc-test-rollout'
--- PASS: TestAccCloudProjectUser_basic (26.24s)
=== RUN   TestAccCloudProjectUser_withRole
    provider_test.go:506: Read Cloud Project /cloud/project/aa2425907763475a89a4bf4e9f9e22fb -> status: 'ok', desc: 'itsc-test-rollout'
--- PASS: TestAccCloudProjectUser_withRole (25.86s)
=== RUN   TestAccCloudProjectUser_withRoles
    provider_test.go:506: Read Cloud Project /cloud/project/aa2425907763475a89a4bf4e9f9e22fb -> status: 'ok', desc: 'itsc-test-rollout'
--- PASS: TestAccCloudProjectUser_withRoles (51.42s)
=== RUN   TestAccCloudProjectUser_withRotate
    provider_test.go:506: Read Cloud Project /cloud/project/aa2425907763475a89a4bf4e9f9e22fb -> status: 'ok', desc: 'itsc-test-rollout'
--- PASS: TestAccCloudProjectUser_withRotate (33.95s)
PASS
ok      github.com/ovh/terraform-provider-ovh/v2/ovh    137.777s
```
Additionally, I have tested the provider in a standard Terraform environment, where all functionality worked as expected. I also verified the migration from provider version 2.2.0 and observed no issues or unexpected behavior.

**Test Configuration**:
* Terraform version: `terraform version`: Terraform v1.11.3
```hcl
resource "ovh_cloud_project_user" "test_user" {
  service_name = var.project_id
  description  = "Test user for password rotation"
  role_name    = "administrator"
  rotate_when_changed = {
    "last_rotated" = "2025-04-30"
  }
}
```

# Checklist:

### DIFF
--- a/ovh/resource_cloud_project_user.go
+++ b/ovh/resource_cloud_project_user.go
@@ -54,6 +54,12 @@ func resourceCloudProjectUser() *schema.Resource {
 				ForceNew: false,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
+			"rotate_when_changed": {
+				Type:        schema.TypeMap,
+				Optional:    true,
+				ForceNew:    false,
+				Description: "A map of arbitrary key/value pairs that will trigger password regeneration when they change, enabling password rotation based on external conditions such as a rotating timestamp.",
+			},
 
 			// Computed
 			"creation_date": {
@@ -159,39 +165,62 @@ func resourceCloudProjectUserUpdate(d *schema.ResourceData, meta interface{}) er
 	config := meta.(*Config)
 	serviceName := d.Get("service_name").(string)
 	userId := d.Id()
-	role := d.Get("role_name")
-	roles, err := helpers.StringsFromSchema(d, "role_names")
-	res := &CloudProjectrolesResponse{}
 
-	res, err = validateCloudProjectUserRoleFunc(config, serviceName, roles, role.(string))
-	if err != nil {
-		return err
-	}
-	update := []string{}
-	for _, i := range res.Roles {
-		if slices.Contains(roles, i.Name) {
-			update = append(update, i.Id)
+	// Check if rotate_when_changed has been modified
+	if d.HasChange("rotate_when_changed") {
+		log.Printf("[DEBUG] rotate_when_changed has been modified, regenerating password for user %s", userId)
+
+		// Call the regeneratePassword endpoint
+		regenerate, err := cloudProjectUserRegeneratePassword(config, serviceName, userId)
+		if err != nil {
+			return fmt.Errorf("error regenerating password for user %s: %s", userId, err)
 		}
-		if role == i.Name && !slices.Contains(update, i.Name) {
-			update = append(update, i.Id)
+
+		// Update the password in the state
+		if err := d.Set("password", regenerate.Password); err != nil {
+			return fmt.Errorf("error setting password in state: %s", err)
+		}
+
+		log.Printf("[DEBUG] Successfully updated password in state for user %s", userId)
+	}
+
+	// Update roles if they changed
+	if d.HasChange("role_name") || d.HasChange("role_names") {
+		role := d.Get("role_name")
+		roles, err := helpers.StringsFromSchema(d, "role_names")
+		res := &CloudProjectrolesResponse{}
+
+		res, err = validateCloudProjectUserRoleFunc(config, serviceName, roles, role.(string))
+		if err != nil {
+			return err
+		}
+		update := []string{}
+		for _, i := range res.Roles {
+			if slices.Contains(roles, i.Name) {
+				update = append(update, i.Id)
+			}
+			if role == i.Name && !slices.Contains(update, i.Name) {
+				update = append(update, i.Id)
+			}
+		}
+
+		log.Printf("[DEBUG] roles IDs %s", update)
+		log.Printf("[DEBUG] user %s", userId)
+		endpoint := fmt.Sprintf("/cloud/project/%s/user/%s/role",
+			url.PathEscape(serviceName),
+			url.PathEscape(userId),
+		)
+
+		r := &CloudProjectUser{}
+		data := &CloudProjectroleUpdate{
+			RolesIds: update,
+		}
+		if err := config.OVHClient.Put(endpoint, data, r); err != nil {
+			return fmt.Errorf("calling %s with params %s:\n\t %q", endpoint, data, err)
 		}
 	}
 
-	log.Printf("[DEBUG] roles IDs %s", update)
-	log.Printf("[DEBUG] user %s", userId)
-	endpoint := fmt.Sprintf("/cloud/project/%s/user/%s/role",
-		url.PathEscape(serviceName),
-		url.PathEscape(userId),
-	)
-
-	r := &CloudProjectUser{}
-	data := &CloudProjectroleUpdate{
-		RolesIds: update,
-	}
-	if err := config.OVHClient.Put(endpoint, data, r); err != nil {
-		return fmt.Errorf("calling %s with params %s:\n\t %q", endpoint, data, err)
-	}
-	return nil
+	return resourceCloudProjectUserRead(d, meta)
 }
 
 func resourceCloudProjectUserCreate(d *schema.ResourceData, meta interface{}) error {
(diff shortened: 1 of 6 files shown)


ovh/resource_cloud_project_user_test.go
```diff
@@ -102,6 +102,91 @@ func TestAccCloudProjectUser_withRoles(t *testing.T) {
...
+			},
+			{
+				Config: testAccCloudProjectUserWithRotateUpdatedConfig,
+				Check: resource.ComposeTestCheckFunc(
+					resource.TestCheckResourceAttr(
+						"ovh_cloud_project_user.user_rotate", "description", "my user for acceptance tests with rotation"),
+					testAccCheckCloudProjectUserOpenRC("ovh_cloud_project_user.user_rotate", t),
+					testAccCheckCloudProjectUserPassword("ovh_cloud_project_user.user_rotate", &secondPassword),
+					testAccCheckCloudProjectUserId("ovh_cloud_project_user.user_rotate", &userId),  // Same user ID (not recreated)
+					testAccCheckCloudProjectUserDifferentPasswords(firstPassword, &secondPassword), // But password changed
```
amstuta, line 151: I just confirmed locally, here `firstPassword` is always uninitialized because this call is made before the test is ran, so it will always be different from `secondPassword`. You should use a pointer for both values so that the check is correct

ovh/resource_cloud_project_user_test.go
```diff
@@ -102,6 +102,91 @@ func TestAccCloudProjectUser_withRoles(t *testing.T) {
...
+resource "ovh_cloud_project_user" "user_rotate" {
+ service_name = "%s"
+ description  = "my user for acceptance tests with rotation"
+ rotate_when_changed = {
+   last_rotation = "2025-04-30"
+ }
+}
+`, os.Getenv("OVH_CLOUD_PROJECT_SERVICE_TEST"))
+
+func TestAccCloudProjectUser_withRotate(t *testing.T) {
```
amstuta, line 125: this test is failing in the end when trying to run post-test destroy because the user is not in status `ok`. This makes me think that the resource should wait for the user to be in `ok` status after the call to `/cloud/project/{serviceName}/user/{userId}/regeneratePassword` to avoid this kind of issue

ovh/resource_cloud_project_user.go
```diff
@@ -54,6 +54,12 @@ func resourceCloudProjectUser() *schema.Resource {
 				ForceNew: false,
 				Elem:     &schema.Schema{Type: schema.TypeString},
 			},
+			"rotate_when_changed": {
+				Type:        schema.TypeMap,
```
amstuta, line 58: see [password_reset](https://registry.terraform.io/providers/ovh/ovh/latest/docs/resources/cloud_project_database_m3db_user#password_reset-1) : we already have this kind of feature on other resources, and we use an arbitrary string. Do you really need a map ? 

I would prefer to keep the same wording/schema between resources.

amstuta (changes requested):
Thanks for your contribution ! I left a few comments but this is a nice feature to add to the provider 👍

ovh/resource_cloud_project_user_test.go
pascalinthecloud: I just did update the tests and did some local testing with debug outputs and seems to work as expected now. :)

ovh/resource_cloud_project_user_test.go
pascalinthecloud: I was able to reproduce the issue (though not every time - sometimes the user returned to ‘ok’ status quickly 😉). Therefore, I added a wait to ensure the user is in ‘ok’ status before proceeding.

ovh/resource_cloud_project_user.go
pascalinthecloud: Thanks, that makes sense! I wasn’t aware there was already a similar feature in other resources. I agree it’s better to stay consistent, so I’ve updated it to use password_reset as a string to match the existing schema.

templates/resources/cloud_project_user.md.tmpl
```diff
@@ -14,6 +14,11 @@ Creates a user in a public cloud project.
 
 {{tffile "examples/resources/cloud_project_user/example_1.tf"}}
 
+{{tffile "examples/resources/cloud_project_user/example_2.tf"}}
+
+-> **NOTE** To reset password of the user previously created, update the `password_reset` attribute. Use the `terraform refresh` command after executing `terraform apply` to update the output with the new password. This attribute can be an arbitrary string but we recommend 2 formats:
```
amstuta, line 19: I think you can remove the part

> Use the `terraform refresh` command after executing `terraform apply` to update the output with the new password

Since password is already updated in the state after a `terraform apply`, no need for the refresh ;)

ovh/resource_cloud_project_user.go
```diff
@@ -159,39 +164,77 @@ func resourceCloudProjectUserUpdate(d *schema.ResourceData, meta interface{}) er
...
+		if err := d.Set("password", regenerate.Password); err != nil {
+			return fmt.Errorf("error setting password in state: %s", err)
+		}
+
+		log.Printf("[DEBUG] Successfully updated password in state for user %s", userId)
+
+		// Wait for user to be ok after password regeneration
+		log.Printf("[DEBUG] Waiting for User %s to be ok after password regeneration:", userId)
+		stateConf := &resource.StateChangeConf{
+			Pending:    []string{"updating", "regenerating"}, // Add any potential intermediate states
```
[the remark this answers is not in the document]
pascalinthecloud: You're completely right, I fixed that.

templates/resources/cloud_project_user.md.tmpl
pascalinthecloud: For the password just to be rotated, the apply is enough, but to apply changes to the output you have to run a refresh after that. 
I just took that notes from the [m3db_user](https://registry.terraform.io/providers/ovh/ovh/latest/docs/resources/cloud_project_database_m3db_user).

amstuta: my bad, I didn't think of the `output` keyword as you don't use it in your example 👍

amstuta (approved):
Thanks for contributing to the provider @pascalinthecloud :)

PR #1082, opened 2025-08-12
add deletion protection field for database service

lpatte:
# Description

Deletion Protection option for cloud Database service

## Type of change

- [X] New feature (non-breaking change which adds functionality)
- [X] Improvement (improve existing resource(s) or datasource(s))
- [X] Documentation update

# How Has This Been Tested?

Tested with this code then try to delete it manualy and by remove the resource to my Terraform code to check that an error is returned and block the deletion.
```hcl
resource "ovh_cloud_project_database" "database" {
  service_name        = "xxxxxxx
  engine              = "postgresql"
  description         = "lucky-yang"
  flavor              = "db1-7"
  nodes {
    region     = "GRA"
  }
  deletion_protection = true
  plan                = "essential"
  version             = "17"
}
```

### DIFF
--- a/docs/data-sources/cloud_project_database.md
+++ b/docs/data-sources/cloud_project_database.md
@@ -55,7 +55,7 @@ The following attributes are exported:
   * `description` - Description of the IP restriction
   * `ip` - Authorized IP
   * `status` - Current status of the IP restriction.
-* `deletion_protection` - Enable delation protection
+* `deletion_protection` - Enable deletion protection
 * `kafka_rest_api` - Defines whether the REST API is enabled on a kafka cluster.
 * `kafka_schema_registry` - Defines whether the schema registry is enabled on a Kafka cluster
 * `maintenance_time` - Time on which maintenances can start every day.
--- a/docs/data-sources/cloud_project_database.md
+++ b/docs/data-sources/cloud_project_database.md
@@ -55,6 +55,7 @@ The following attributes are exported:
   * `description` - Description of the IP restriction
   * `ip` - Authorized IP
   * `status` - Current status of the IP restriction.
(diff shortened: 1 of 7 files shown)


docs/data-sources/cloud_project_database.md
```diff
@@ -55,6 +55,7 @@ The following attributes are exported:
   * `description` - Description of the IP restriction
   * `ip` - Authorized IP
   * `status` - Current status of the IP restriction.
+* `deletion_protection` - Enable delation protection
```
amstuta:
```suggestion
* `deletion_protection` - Enable deletion protection
```

amstuta: or even `Status of deletion protection`

ovh/data_cloud_project_database.go
```diff
@@ -130,6 +130,11 @@ func dataSourceCloudProjectDatabase() *schema.Resource {
 					},
 				},
 			},
+			"deletion_protection": {
+				Type:        schema.TypeBool,
+				Description: "Enable deletion protection",
+				Optional:    true,
```
amstuta:
```suggestion
				Computed:    true,
```

ovh/resource_cloud_project_database.go
```diff
@@ -77,6 +77,11 @@ func resourceCloudProjectDatabase() *schema.Resource {
 					},
 				},
 			},
+			"deletion_protection": {
```
amstuta, line 80: should also have `Computed: true` right ?

lpatte: It's not computed in resource, it's an optional flag, user can set like kafka_rest_api or opensearch_acls_enabled

PR #1162, opened 2025-11-20
fix(okms_secret): detect drift on data change

### LINKED ISSUE #1147: [BUG] No data change recognition in ovh_okms_secret
### Describe the bug
There is a synchronization issue between the secret resource and terraform. In most resources (e.g. storage) some changed values in UI side will be recongnized and correctly shown.

My use case is following:
Create a secret resource with default values in terraform and manually fill in the real secret value as a person. That way there is no secret in the code. In my example is no prevention of changes to the secret data by adding a lifecycle rule.

But in any case I would like to get the real value from my terraform provider. I found a tedious workaround, but that is unintuitive and is a lot of unncessary extra code.

### Terraform Version

Terraform v1.13.4
on darwin_arm64

provider registry.terraform.io/ovh/ovh v2.9.0

### Affected Resource(s)

Please list the resources as a list, for example:
- ovh_okms_secret

### Terraform Configuration Files

```hcl
resource "ovh_okms_secret" "my_credentials" {
  okms_id = var.ovh_secret_manager
  path    = "my_credentials"

  version = {
    data = jsonencode({
      password = "MyPassword"
    })
  }
}

# resource does not contain actual values from secret store
data "ovh_okms_secret" "my_credentials" {
  okms_id      = var.ovh_secret_manager
  path         = ovh_okms_secret.my_credentials.path
  include_data = true
}

output "my_buggy_credentials" {
  value     = ovh_okms_secret.my_credentials.version.data
  sensitive = true
}

output "my_real_credentials" {
  value     = data.ovh_okms_secret.my_credentials.data
  sensitive = true
}
```

### Steps to Reproduce

Please list the steps required to reproduce the issue, for example:
1. `terraform apply`
2. Go to secret manager and change the password to a new value
3. `terraform apply` again
4. `terraform output my_buggy_credentials`
5. `terraform output my_real_credentials`

### Debug Output

For step 4
```
"{\"password\":\"MyPassword\"}"
```

For step 5
```
"{\"password\":\"MyChangedPassword\"}"
```

### Expected Behavior

1. Terraform should have actually recognized that the resource is not correct and changed the value back to my password.
2. Terraform should not yield different values for the same entity. Resource and data should be equal.

alexGNX:
# Description

The fix is about detecting changes made to an `okms_secret` resource outside of terraform and reconcile the states.

Fixes #1147  (issue)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

```
export TF_ACC=1
export OVH_ENDPOINT=ovh-eu
export OVH_APPLICATION_KEY=...
export OVH_APPLICATION_SECRET=...
export OVH_CONSUMER_KEY=...
export OVH_OKMS=...

# data source tests
go test -timeout 30s -run ^TestAccOkmsSecretDataSource_latestAndVersions$ github.com/ovh/terraform-provider-ovh/v2/ovh
go test -timeout 60s -run ^TestAccOkmsSecretDataSource_latestAndVersions$ github.com/ovh/terraform-provider-ovh/v2/ovh

# resource tests
go test -timeout 90s -run ^TestAccOkmsSecretResource_basicLifecycle$ github.com/ovh/terraform-provider-ovh/v2/ovh
go test -timeout 5m -run TestAccOkmsSecretResource_basicLifecycle github.com/ovh/terraform-provider-ovh/v2/ovh
go test -timeout 5m -run TestAccOkmsSecretResource_casMismatch github.com/ovh/terraform-provider-ovh/v2/ovh
go test -timeout 5m -run TestAccOkmsSecretResource_pathRecreateAndMetadata github.com/ovh/terraform-provider-ovh/v2/ovh
```

# Checklist:

### DIFF
--- a/ovh/resource_okms_secret.go
+++ b/ovh/resource_okms_secret.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/url"
+	"reflect"
+	"slices"
 
 	"github.com/hashicorp/terraform-plugin-framework/resource"
 	ovhtypes "github.com/ovh/terraform-provider-ovh/v2/ovh/types"
@@ -82,17 +84,17 @@ func (r *okmsSecretResource) Create(ctx context.Context, req resource.CreateRequ
 }
 
 func (r *okmsSecretResource) Read(ctx context.Context, req resource.ReadRequest, resp *resource.ReadResponse) {
-	var data, responseData OkmsSecretModel
+	var secretFromState, currentSecret OkmsSecretModel
 
 	// Read Terraform prior state data into the model
-	resp.Diagnostics.Append(req.State.Get(ctx, &data)...)
+	resp.Diagnostics.Append(req.State.Get(ctx, &secretFromState)...)
 	if resp.Diagnostics.HasError() {
 		return
 	}
 
-	endpoint := "/v2/okms/resource/" + url.PathEscape(data.OkmsId.ValueString()) + "/secret/" + url.PathEscape(data.Path.ValueString()) + ""
+	endpoint := "/v2/okms/resource/" + url.PathEscape(secretFromState.OkmsId.ValueString()) + "/secret/" + url.PathEscape(secretFromState.Path.ValueString())
 
-	if err := r.config.OVHClient.Get(endpoint, &responseData); err != nil {
+	if err := r.config.OVHClient.Get(endpoint, &currentSecret); err != nil {
 		resp.Diagnostics.AddError(
 			fmt.Sprintf("Error calling Get %s", endpoint),
 			err.Error(),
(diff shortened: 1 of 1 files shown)


ovh/resource_okms_secret.go
```diff
@@ -244,6 +291,34 @@ func buildMetadataPayload(meta *MetadataValue) map[string]any {
...
+		return a == b
+	}
+	normalizedB, err := json.Marshal(objB)
+	if err != nil {
+		return a == b
+	}
+
+	return string(normalizedA) == string(normalizedB)
+}
+
```
jrm-d: Maybe we could reduce the amount of []byte <--> string conversions here, for example:
```
func semanticJSONEqual(a, b []byte) bool {
    var objA, objB any

    // Try to parse both as JSON
    if err := json.Unmarshal(a, &objA); err != nil {
        // Not valid JSON, fall back to []byte comparison
        return slices.Equal(a, b)
    }
    if err := json.Unmarshal(b, &objB); err != nil {
        return false
    }

    return reflect.DeepEqual(objA, objB)
} 
```
you would only have to convert input parameters as []byte to call this func

jrm-d (changes requested):

ovh/resource_okms_secret.go
alexGNX: done, thanks for the suggestion.

jrm-d (approved):

Issue #1115, opened 2025-09-15
[FEATURE] Create public cloud instances with multiple network interfaces


InversMoritz:
### Description

It would be great if it would be possible to create public cloud instances with multiple network interfaces

### Affected Resource(s) and/or Data Source(s)

* ovh_cloud_project_instance

### Potential Terraform Configuration

Maybe just with multiple `network`-blocks in instance.network.private?

```
resource "ovh_cloud_project_instance" "instance" {
  service_name  = "XXX"
    region = "RRRR"
    billing_period = "hourly"
    boot_from {
        image_id = "UUID"
    }
    flavor {
        flavor_id = "UUID"
    }
    name = "instance name"
    ssh_key {
        name = "sshname"
    }
    network {
        public = false
        private = {
          network = {
            id = var.network_id_1
            subnet_id = var.subnet_id_1
          }
          network = {
            id = var.network_id_2
            subnet_id = var.subnet_id_2
          }
        }
    }  
}
```

### Additional context

Add any other context, HCL configuration or screenshots about the feature request here.

PR #1164, opened 2025-11-25
Dev/thekrauss/cloud instances with multiple network

thekrauss:
Description

This PR addresses issue #1115 by enabling the attachment of multiple private network interfaces to a Public Cloud instance.

Previously, the provider limited the network.private block to a single item (MaxItems: 1). This change allows users to define multiple private blocks to connect an instance to multiple vRacks/Subnets simultaneously (e.g., Front/Back architecture).

## Type of change

• Schema: Changed network.private from TypeSet to TypeList to ensure interface ordering (eth1, eth2...) is respected and removed the MaxItems: 1 restriction.

• Structures: Updated Network struct to handle a slice of private networks ([]*PrivateNetwork) instead of a single pointer.

• Logic: Updated GetNetwork helper to iterate over all defined private blocks and construct the correct payload for the API

```hcl
resource "ovh_cloud_project_instance" "multinic_instance" {
  service_name   = "XXX"
  region         = "GRA11"
  name           = "multi-nic-instance"
  flavor_id      = "UUID"
  image_id       = "UUID"

  network {
    public = true

    private {
      network {
        id        = private.net_front.id
        subnet_id = subnet.subnet_front.id
      }
      ip = "10.0.1.10"
    }

    private {
      network {
        id        = private.net_back.id
        subnet_id = subnet.subnet_back.id
      }
    }
  }
}

```

### DIFF
--- a/env.sh
+++ b/env.sh
@@ -1,12 +1,3 @@
 #!/usr/bin/env bash
 
-export OVH_ENDPOINT="ovh-eu"    
-export OVH_APPLICATION_KEY="000693a9655db1b5"
-export OVH_APPLICATION_SECRET="dda3dcae9b4fbd41461c07d100da6586"
-export OVH_CONSUMER_KEY="055d118f9c3ca88b4fba9ae1b984fdf8"
-
-export OVH_DOMAIN_TEST="syk-edu.com"
-
-export TF_ACC=1
-
-echo "[OK] OVH environment variables exported for acceptance tests."
+!!!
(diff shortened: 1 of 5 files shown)


ShinySaana:
This PR would have been very welcome for our use case. Thank you for the work that was put in making it.

amstuta:
@ShinySaana we understand that this feature is really expected, and we are currently working on a new AP to expose the instances, which will allow setting multiple networks. It should be available in the next few months, and as soon as the API is available we'll add the feature in the Terraform provider.

PR #1178, opened 2025-12-04
fix: populate ID field in ovh_cloud_project_storage resource

fabiencastarede:
## Description

Fixes the `ovh_cloud_project_storage` resource to properly populate the `id` field in Terraform state.

## Problem

The `ovh_cloud_project_storage` resource declares an `id` attribute in its schema but never populates it, leaving it as `null` in the Terraform state. According to [Terraform Plugin Framework
best practices](https://developer.hashicorp.com/terraform/plugin/framework/resources/state#id-attribute), if an `id` attribute is declared in the schema, it should be populated.

This causes issues with:
- Tools and frameworks that expect a populated ID field in Terraform state
- Inconsistency with other resources in the provider (e.g., `ovh_cloud_project_database` properly populates its ID)
- User expectations when inspecting state files

## Changes

- Add `ovhtypes` import for TfString type creation
- Populate ID in `Create()` function after resource creation
- Populate ID in `Read()` function after fetching resource details
- Populate ID in `Update()` function after updating resource
- ID format: `service_name/region_name/name` (matches the import format defined in `ImportState()`)

## Benefits

- Complies with Plugin Framework best practices
- Provides consistency across the provider
- Makes the ID format explicit and matches import expectations
- Improves tooling compatibility

## Testing

Tested with:
- Resource creation ✅
- Resource updates ✅
- Resource deletion ✅
- Resource import ✅
- State inspection ✅

The existing acceptance tests continue to pass as they already use the correct import ID format (`service_name/region_name/name`).

## Backward Compatibility

This change is backward compatible since the ID field was previously `null`. Adding a value doesn't break existing workflows.

### DIFF
--- a/ovh/resource_cloud_project_storage_test.go
+++ b/ovh/resource_cloud_project_storage_test.go
@@ -11,6 +11,7 @@ import (
 
 func TestAccCloudProjectRegionStorage_basic(t *testing.T) {
 	bucketName := acctest.RandomWithPrefix(test_prefix)
+	serviceName := os.Getenv("OVH_CLOUD_PROJECT_SERVICE_TEST")
 	config := fmt.Sprintf(`
 	resource "ovh_cloud_project_storage" "storage" {
 		service_name = "%s"
@@ -20,7 +21,7 @@ func TestAccCloudProjectRegionStorage_basic(t *testing.T) {
 			status = "enabled"
 		}
 	}
-	`, os.Getenv("OVH_CLOUD_PROJECT_SERVICE_TEST"), bucketName)
+	`, serviceName, bucketName)
 
 	resource.Test(t, resource.TestCase{
 		PreCheck:                 func() { testAccPreCheckCloud(t); testAccCheckCloudProjectExists(t) },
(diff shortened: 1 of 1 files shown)


ovh/resource_cloud_project_storage.go
```diff
@@ -118,9 +125,12 @@ func (r *cloudProjectStorageResource) Read(ctx context.Context, req resource.Rea
 
 	responseData.MergeWith(&data)
 
-	if data.HideObjects.ValueBool() {
```
ekarlso, line 128: You should probably not remove this

ovh/resource_cloud_project_storage.go
```diff
@@ -175,9 +185,12 @@ func (r *cloudProjectStorageResource) Update(ctx context.Context, req resource.U
 
 	responseData.MergeWith(&planData)
 
-	if planData.HideObjects.ValueBool() {
```
ekarlso, line 185: You should probably not remove this

ovh/resource_cloud_project_storage.go
fabiencastarede: Thanks, the master branch has evolved while I was making my PR...
Now It's rebased.

ovh/resource_cloud_project_storage.go
fabiencastarede: Thanks, the master branch has evolved while I was making my PR...
Now It's rebased.

jrm-d (approved):

PR #1216, opened 2026-01-26
feat(ip_move): remove useless calls to /ip/service

tduboys:
# Description

This PR removes useless calls to `/ip/service/{service_name}` when moving an additionnal IP, as it's not mandatory to manage an additional IP, and in some cases, the service name is not predictible and the current helper return a wrong value.

The call `/ip/{ip}/move` also have an extra parameter called `nexthop` that is useful for PCC or PCI routing, but the call `/ip/{ip}` is not able to provide the current value of this flag, so it's better to not use it on Terraform, and uses a dedicated resource like `move_ip_to_cloud_project`

## Type of change

- [x] Documentation update

# Checklist:

### DIFF
--- a/ovh/resource_ip_move.go
+++ b/ovh/resource_ip_move.go
@@ -10,7 +10,6 @@ import (
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
-	"github.com/hashicorp/terraform-plugin-testing/helper/resource"
 	"github.com/ovh/go-ovh/ovh"
 	"github.com/ovh/terraform-provider-ovh/v2/ovh/helpers"
 	"golang.org/x/exp/slices"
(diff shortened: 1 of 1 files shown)


ovh/resource_ip_move.go
```diff
@@ -137,23 +154,16 @@ func resourceIpMoveUpdate(d *schema.ResourceData, meta interface{}) error {
 		return err
 	}
 
-	serviceName, err := helpers.ServiceNameFromIpBlock(ip)
```
amstuta, line 140: I know that `serviceName` can contain an invalid value in some cases, but could you keep it valued for now ? Otherwise it is a breaking change because customers may rely on this value right now.
IMO we should add an issue with the `Breaking Change` label to completely remove this field in the next major release.

PR #1228, opened 2026-02-10
feat(storage): add EFS service resource

dariakts:
# Description

Add EFS service resource.

## Type of change

- [x ] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

- [x ] Test EFS service order: `make testacc TESTARGS="-run TestAccResourceStorageEfs_basic"`
- [x ] Test EFS service order with name update: `make testacc TESTARGS="-run TestAccResourceStorageEfs_updateName"`
- [x ] Test EFS service import:  `make testacc TESTARGS="-run TestAccResourceStorageEfs_import"`

**Test Configuration**:
* Terraform version: `terraform version`: Terraform v1.8.3
```hcl
data "ovh_me" "my_account" {}

resource "ovh_storage_efs" "efs" {
  ovh_subsidiary = data.ovh_me.my_account.ovh_subsidiary

  plan = [
    {
      plan_code = "enterprise-file-storage-premium-1tb"
      duration = "P1M"
      pricing_mode = "default"

      configuration = [
        {
          label = "region"
          value = "eu-west-lim"
	},
	{
	  label = "network"
	  value = "vrack"
	}
      ]
    }
  ]
}
```

# Checklist:


amstuta:
## Summary

This PR adds a new Terraform resource `ovh_storage_efs` to order and manage an OVHcloud Enterprise File Storage (EFS / NetApp) service. It follows the order-based resource pattern already used elsewhere in the provider.

---

## 1. Compilation Issues

### 1.1 LIKELY OK — Type and import usage

The resource uses types (`StorageEfsResourceModel`, `PlanValue`, `OrderModel`, `IamValue`, `OrderValue`, `ConfirmTerminationOpts`, etc.) and functions (`orderCreate`, `serviceNameFromOrder`, `orderDelete`, `OrderResourceSchema`) that are defined elsewhere in the `ovh` package. The imports reference `ovhtypes "github.com/ovh/terraform-provider-ovh/v2/ovh/types"` and the standard Terraform Framework libraries. These are consistent with existing resources like `resource_storage_efs_share.go`.

**Verdict:** No compilation issues detected from the diff alone. The types and helpers referenced are consistent with existing codebase patterns.

---

## 2. Terraform Logic Issues

### 2.1 CRITICAL — Missing `return` after errors in `Create`

**File:** `resource_storage_efs.go`, lines ~80-95 of the new file

```go
// Create order and wait for service to be delivered
order := data.ToOrder()
if err := orderCreate(order, r.config, "netapp", true, defaultOrderTimeout); err != nil {
    resp.Diagnostics.AddError("failed to create order", err.Error())
    // ❌ MISSING: return
}

// ...

serviceName, err := serviceNameFromOrder(r.config.OVHClient, orderID, plans[0].PlanCode.ValueString())
if err != nil {
    resp.Diagnostics.AddError("failed to retrieve service name", err.Error())
    // ❌ MISSING: return
}
```

**Impact:** If `orderCreate` fails, the code continues to access `order.Order.OrderId` which could be zero-valued or cause a nil-pointer panic. Similarly, if `serviceNameFromOrder` fails, the code proceeds to use an empty `serviceName` for API calls, resulting in confusing secondary errors or incorrect API calls (`PUT /storage/netapp/`).

**Fix:** Add `return` after each `resp.Diagnostics.AddError(...)` call, consistent with every other error-handling block in the same file (and the share resource).

### 2.2 MEDIUM — No wait/poll for service readiness after order delivery

In `Create`, after the order is placed and the service name retrieved, the code immediately calls `PUT /storage/netapp/{serviceName}` to update the name. However, the service might still be in `creating` status at that point. The `orderCreate` function waits for order completion, but order completion doesn't necessarily mean the NetApp service is in `running` state.

By contrast, the share resource has `WaitForShareCreation` that polls until the status is `available`. No equivalent polling exists here for the EFS service status.

**Risk:** The PUT call could fail with a 409 or similar if the service isn't fully ready yet. This may be mitigated by `orderCreate`'s `waitForCompletion=true` parameter, but it depends on whether order completion guarantees service readiness.

**Recommendation:** Consider adding a polling loop that waits for `status == "running"` before calling PUT.

### 2.3 LOW — Unnecessary empty string concatenation

**File:** `resource_storage_efs.go`, multiple locations

```go
endpoint := "/storage/netapp/" + url.PathEscape(data.ServiceName.ValueString()) + ""
```

The trailing `+ ""` is a no-op. It appears in Create (twice) and Read. The Update method doesn't have it. This is cosmetic but inconsistent.

### 2.4 LOW — Update MergeWith ordering

In `Update`:
```go
responseData.MergeWith(&planData)
responseData.MergeWith(&data)
```

This fills API response gaps with plan data first, then prior state data. This is correct — API response takes priority, then planned values, then prior state. Consistent with the share resource pattern.

---

## 3. Schema vs API Spec

### API Reference: `storage.NetAppServiceWithIAM` (GET response)

| API Field | API Type | Schema Attribute | Schema Config | Match? |
|-----------|----------|------------------|---------------|--------|
| `createdAt` | string (date-time, readOnly) | `created_at` | Computed | ✅ |
| `id` | string (UUID, readOnly) | `id` | Computed | ✅ |
| `name` | string (readOnly on GET) | `name` | Optional | ✅ (see note) |
| `performanceLevel` | enum [premium] (readOnly) | `performance_level` | Computed | ✅ |
| `product` | string (readOnly) | `product` | Computed | ✅ |
| `quota` | integer (readOnly) | `quota` | Computed (Int64) | ✅ |
| `region` | enum (readOnly) | `region` | Computed | ✅ |
| `status` | enum (readOnly) | `status` | Computed | ✅ |
| `iam` | ResourceMetadata (readOnly) | `iam` | Computed (nested) | ✅ |

### API Reference: `storage.NetAppServiceUpdate` (PUT request body)

| API Field | API Type | Mapped from | Match? |
|-----------|----------|-------------|--------|
| `name` | string (**required**) | `StorageEfsResourceWritableModel.Name` | ⚠️ See 3.1 |

### 3.1 MEDIUM — `name` is required in API PUT but Optional in schema

The API spec says `name` is **required** in `storage.NetAppServiceUpdate`. However:
- The schema marks `name` as `Optional` only (not Required)
- `ToUpdate()` uses `json:"name,omitempty"` with a pointer, meaning if `name` is not set, it's omitted from the JSON body
- The PUT is called unconditionally in `Create` even when `name` is not provided

**Risk:** If a user doesn't specify `name`, the Create will send `PUT /storage/netapp/{id}` with body `{}`, which may fail because the API requires `name`. Similarly, calling Update without changing `name` should still include it.

**Recommendation:** Either:
- Make `name` Required in the schema, or
- Add `Computed: true` alongside `Optional: true` and handle the case where the API auto-generates a name, or
- Skip the PUT call in Create when `name` is not set

### 3.2 LOW — `name` should likely also be `Computed`

The `name` field is `Optional` but not `Computed`. When importing an existing EFS or when a name is auto-assigned by the service, the Read function will populate `name` from the API response via `MergeWith`. Without `Computed: true`, the Terraform Framework may produce inconsistent plan output (reporting a value for `name` that wasn't in the config).

**Recommendation:** Add `Computed: true` to the `name` attribute to handle imports and API-assigned names correctly.

### 3.3 INFO — No enum validation on computed fields

Fields like `performance_level` (enum: `premium`), `region` (enum: `ca-east-bhs`, `eu-west-gra`, etc.), and `status` (enum: `creating`, `running`, etc.) don't have validators. This is acceptable for Computed-only fields since the values come from the API, but worth noting for documentation accuracy.

### 3.4 OK — IAM sub-schema

The IAM nested schema matches `iam.ResourceMetadata`:
- `display_name` → `displayName` ✅
- `id` → `id` ✅
- `tags` → `tags` (Map of String) ✅
- `urn` → `urn` ✅

Note: The API has a `state` field in `iam.ResourceMetadata` (enum: `EXPIRED`, `IN_CREATION`, `OK`, `SUSPENDED`) that is **not** in the schema. This is consistent with other resources in the provider (the `state` field is typically omitted from IAM sub-schemas), so this is fine.

### 3.5 OK — Order-related attributes

The order attributes (`ovh_subsidiary`, `plan`, `plan_option`, `order`) come from `OrderResourceSchema(ctx)` and are reused across multiple order-based resources. They are not specific to this PR and assumed correct.

---

## 4. State Management

### 4.1 GOOD — Early state save in Create

```go
// Save early data into Terraform state, to make sure a reapply will not reorder a new EFS
resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)
```

This is an important pattern for order-based resources: if the subsequent PUT or GET fails, the service name is already in state, preventing a duplicate order on the next `terraform apply`. This is correct.

### 4.2 GOOD — Import sets both `service_name` and `id`

```go
func (r *storageEfsResource) ImportState(ctx context.Context, req resource.ImportStateRequest, resp *resource.ImportStateResponse) {
    resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("service_name"), req.ID)...)
    resp.Diagnostics.Append(resp.State.SetAttribute(ctx, path.Root("id"), req.ID)...)
}
```

This correctly sets both attributes since `id` and `service_name` are the same value. The subsequent Read will use `service_name` to call the API and populate all computed fields.

### 4.3 GOOD — Delete handles 404/460 gracefully

```go
if errOvh, ok := err.(*ovh.APIError); ok && (errOvh.Code == 404 || errOvh.Code == 460) {
    return "", nil
}
```

The terminate function gracefully handles already-deleted services (404) and gone services (460).

### 4.4 MEDIUM — Create doesn't set `id` before early state save

In Create, the early state save happens before `data.Id = data.ServiceName` is set. The `id` is only set after the final GET:

```go
// Save early data into Terraform state  ← id is still unknown here
resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)

// ... PUT, GET ...

data.Id = data.ServiceName  // ← id set here
resp.Diagnostics.Append(resp.State.Set(ctx, &data)...)  // ← final save
```

If the function fails between the early save and the final save, the state will have `service_name` set but `id` as unknown/null. Since `id` is used as the Terraform resource identifier, this could cause issues on subsequent operations (terraform may not be able to find the resource).

**Recommendation:** Set `data.Id = data.ServiceName` before the early state save.

### 4.5 OK — MergeWith logic

The `MergeWith` function correctly uses the pattern:
```go
if (v.Field.IsUnknown() || v.Field.IsNull()) && !other.Field.IsUnknown() {
    v.Field = other.Field
}
```

This ensures existing non-null values are preserved and only gaps are filled, which is the correct behavior for merging API responses with plan/state data.

---

## 5. Documentation & Examples

### 5.1 BUG — Example HCL references wrong data source name

**Files:** `docs/resources/storage_efs.md`, `examples/resources/storage_efs/resource.tf`

```hcl
data "ovh_me" "my_account" {}

resource "ovh_storage_efs" "efs" {
  ovh_subsidiary = data.ovh_me.myaccount.ovh_subsidiary
                                ^^^^^^^^^ should be "my_account"
```

The data source is declared as `"my_account"` but referenced as `myaccount` (missing underscore). This would cause a Terraform plan error.

### 5.2 INFO — Pre-existing typo

In `docs/index.md` and `templates/index.md.tmpl`:
```
The service name of your Enteprise File Storage service
                          ^^^^^^^^^
```
"Enteprise" should be "Enterprise". This typo pre-exists in the codebase (not introduced by this PR).

### 5.3 OK — Test coverage

The test (`resource_storage_efs_test.go`) covers:
- Basic creation with plan configuration
- Verifies computed attributes are set (`created_at`, `iam.urn`, `id`, `name`, `performance_level`, `product`, `quota`)
- Verifies specific values (`region`, `status`)
- Uses proper pre-check (`testAccPreCheckOrderStorageEfs`)
- Uses environment variable for region

Missing: No update test (e.g., changing `name`) and no import test. These would be valuable additions.

---

## 6. Summary of Findings

| # | Severity | Category | Description |
|---|----------|----------|-------------|
| 2.1 | **CRITICAL** | Logic | Missing `return` after `AddError` in Create (2 locations) — can cause panics or cascading errors |
| 5.1 | **BUG** | Docs | Example references `myaccount` instead of `my_account` |
| 3.1 | **MEDIUM** | Schema/API | `name` is required in API PUT but Optional in schema; PUT called unconditionally |
| 3.2 | **MEDIUM** | Schema | `name` should be `Computed: true` alongside `Optional` for imports/API defaults |
| 2.2 | **MEDIUM** | Logic | No polling for service readiness before PUT in Create |
| 4.4 | **MEDIUM** | State | `id` not set before early state save |
| 2.3 | **LOW** | Style | Unnecessary `+ ""` in endpoint strings |
| 3.3 | **INFO** | Schema | No enum validation on computed fields (acceptable) |
| 5.3 | **INFO** | Tests | No update or import test cases |

PR #1353, opened 2026-06-19
fix(docs): correct URN resource type in iam_resource_tags examples

pascalinthecloud:
## Summary

- Fixes the URN format in the `ovh_iam_resource_tags` resource documentation
- Replaces `cloudProject` with `publicCloudProject` as the resource type in example URNs
- Affected files: example TF file, template, and generated docs

## Test plan

- [x] Verify the [registry docs page](https://registry.terraform.io/providers/ovh/ovh/latest/docs/resources/iam_resource_tags) shows the correct URN format after release


amstuta (approved):
Thanks for your PR @pascalinthecloud !

